PR #32, opened 2023-04-14
24 package care

Tyroonis:
Added packages:
"com.unity.xr.interaction.toolkit": "2.1.1",
"com.unity.xr.management": "4.2.1",
"com.unity.xr.openxr": "1.5.3",

Added plugins:
Pico Open XR plugin

- Test framework was already activated
- no updates needed


JaXt0r:
Stupid question: Did you copy&paste from previous project or did you Install packages via package Manager? The latter would be awesome as it would automatically install latest versions.

JaXt0r:
Also: Did you test already if you could walk in VR? St least via mockBRPlayer (forgot name of this feature where you can simulate VR gestures via Mouse+keyboard).
Would be great to have a quick test before merging. Just tell me and I could also do the test.

Tyroonis:
I installed the packages via the package manager.
I have downloaded the Pico SDK again.

I did not do a VR.walk.
But I looked if the sdks were recognized and the corresponding builds work as well.

On the PC build it launches Steam VR and on the glasses they open in VR mode.

JaXt0r (approved):
Thanks for your feedback. Sounds good. Approved.

PR #42, opened 2023-04-18
Feature/android settings import

JucanAndreiDaniel:


Assets/unZENity-VR/Scripts/Settings/SettingsManager.cs
```diff
@@ -23,19 +23,26 @@ private void LoadGameSettings()
         {
             var settingsFilePath = $"{GetRootPath()}/{SETTINGS_FILE_NAME}";
             if (!File.Exists(settingsFilePath))
-                throw new ArgumentException($"No >GameSettings.json< file exists at >{settingsFilePath}<. Can't load Gothic1.");
+                if (Application.platform == RuntimePlatform.Android)
+                    Importer();
```
JaXt0r, line 27: We still need to throw an exception.

Because by default our game doesn't know where to look for the Gothic installation.
Therefore let's throw the error and tell user "what to do next".

Assets/unZENity-VR/Scripts/Settings/SettingsManager.cs
```diff
@@ -65,5 +72,28 @@ private void CheckIfGothicIDirectoryExists()
                     $"Please put in the right absolute path to your local installation.");
         }
 
+        /// <summary>
+        /// Import the settings file from streamingAssetPath to persistentDataPath.
+        /// Since the settings file is in streamingAssetPath, we need to use UnityWebRequest to move it so we can have access to it
+        /// as detailed here https://docs.unity3d.com/ScriptReference/Application-streamingAssetsPath.html
+        /// </summary>
+        private void Importer()
```
JaXt0r, line 80: Can you rename this method, please. I think of something which tells us "It's cloning the GameSettings.json file to be altered for next game start".

How about sth. like _CopyGameSettingsForAndroidBuild()_?

Assets/unZENity-VR/Scripts/Settings/SettingsManager.cs
```diff
@@ -65,5 +72,28 @@ private void CheckIfGothicIDirectoryExists()
...
+        {
+            string GameSettingsPath = System.IO.Path.Combine(Application.streamingAssetsPath, $"{SETTINGS_FILE_NAME}");
+            string result = "";
+            if (GameSettingsPath.Contains("://") || GameSettingsPath.Contains(":///"))
```
JaXt0r, line 84: I think it will either always be true or always be false. 🤔

Assets/unZENity-VR/Scripts/Settings/SettingsManager.cs
```diff
@@ -65,5 +72,28 @@ private void CheckIfGothicIDirectoryExists()
...
+            {
+                UnityEngine.Networking.UnityWebRequest www = UnityEngine.Networking.UnityWebRequest.Get(GameSettingsPath);
```
JaXt0r, line 86: Do we really need WebRequest? Can't we just leverage default System.Io.File.Copy() or sth. like that?
(I don't know if this can be used to copy data from within an APK into the local file system, but we should test.)

JaXt0r (changes requested):

JaXt0r:
Good first PR. A few hints only. :-)

Assets/unZENity-VR/Scripts/Settings/SettingsManager.cs
JucanAndreiDaniel: `Android uses a compressed .apk file. These platforms return a URL. Use the [UnityWebRequest](https://docs.unity3d.com/ScriptReference/Networking.UnityWebRequest.html) class to access the Assets.` this is a quote directly from https://docs.unity3d.com/ScriptReference/Application-streamingAssetsPath.html

Assets/unZENity-VR/Scripts/Settings/SettingsManager.cs
JucanAndreiDaniel: It's just a check to see if we have a URL, or a path from a jar (ex `"jar:file://`) if not just copy from one place to another

Assets/unZENity-VR/Scripts/Settings/SettingsManager.cs
JucanAndreiDaniel: I'll rename it 👍

Assets/unZENity-VR/Scripts/Settings/SettingsManager.cs
JaXt0r: I understand now: If the file doesn't exist, just create it. True!

PR #45, opened 2023-04-19
add EditorTests for FilesCheck

Tyroonis:
First tests added

At the moment we are only checking if the libraries are present.


Tyroonis:
Checks that have been run can be viewed afterwards.


image

JaXt0r (changes requested):
Nice nice. Thanks for starting tests before merge. This will help us "not accidentaly doing unwanted things".

Some improvement ideas:

You're doing an additional test method for every file to check. This adds about 10 lines of duplicate lines for each check.

I would recommend you make the tests more dynamic like:
* Have a .txt/.config/.yml file where you name every file to check in a separate line
* Have a loop in your "FileExistsTest()" and assert if every line item exists.

Justification:
* Adding a file to check only requires one line in a configuration file which makes it easy to adopt.

Additionally:
* Please put the tests inside /Assets/unZENity-VR/EditorTests (the unZENity-VR folder shows us, that the tests are from us, not another package like OpenXR etc.)
* Cool, that you're leveraging official C# framework NUnit. On PxCs project I used XUnit as it was named the "modern" alternative. Can you briefly check if it makes sense for you to switch (or myself to switch on PxCs. 😉)

Tyroonis:
@JaXt0r 
I deliberately did not put the tests in the project.
It does not belong directly to the project.

Tyroonis:
I have reduced the dll checks to 1 file and moved the tests to ProjectFolder.

XUnit is not supported by Unity. 
Therefore NUnit.

Assets/unZENity-VR/Tests/EditorTests/FilesCheck/FilesCheck.asmdef
```diff
@@ -0,0 +1,9 @@
+{
```
JaXt0r, line 1: What is this file? Just for my interest. :-)

Assets/unZENity-VR/Tests/EditorTests/FilesCheck/dll-List.txt
```diff
@@ -0,0 +1,6 @@
+I18N.West.dll
```
JaXt0r, line 1: Nice!

Assets/unZENity-VR/Tests/EditorTests/FilesCheck/Libraries_exists.cs
JaXt0r, line 1: One additional hint (But can also be for a second PR:) Can we please check if the libphoenix-shared:* files have updatedat with a time difference <3 days or so? This will really help us to always (e.g.) commit the arm .so together with .dll (If we forget, we will get hard to debug errors on pico/quest.

@see: https://stackoverflow.com/a/12484575/3680249

```c#
  var info = new FileInfo(@"C:\thefile.txt");    
  var created = info.CreationTime;    
  var modified = info.LastWriteTime;
   ```

JaXt0r (changes requested):
One open question about timestamps.

Assets/unZENity-VR/Tests/EditorTests/FilesCheck/FilesCheck.asmdef
Tyroonis: Only an assembly reference for the Testrunner.
So that it knows where the tests are.

Tyroonis:
- Merged Main into PR
- Tests adjusted

JaXt0r (approved):

PR #46, opened 2023-04-20
Fix NPC names for gameobjects

JucanAndreiDaniel:
Fixes NPC names not being set for gameobjects making debugging easier in the future.


JaXt0r:
Ah cool. Thanks for this fine little fix.

PR #56, opened 2023-04-25
55 pipeline not for changes in docs

Tyroonis:
Pipeline now ignores changes in docs and the readme

### DIFF
--- a/.github/workflows/test_and_build.yml
+++ b/.github/workflows/test_and_build.yml
@@ -6,14 +6,14 @@ on:
       - main
     types: [opened, reopened, synchronize]
     paths-ignore:
-      - 'docs/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/test_and_build.yml
```diff
@@ -5,10 +5,16 @@ on:
     branches:
       - main
     types: [opened, reopened, synchronize]
+    paths-ignore:
+      - 'docs/**'
```
JaXt0r, line 9: I already suggested if we name docs to Docs (naming convention wide).

So is this case insensitive?

JaXt0r (approved):

.github/workflows/test_and_build.yml
Tyroonis: fixed

PR #72, opened 2023-05-04
71 controller standalone not working

Tyroonis:
- "old" Oculus controller added again
- deprecated using directives replaced


JaXt0r (approved):
Small changes. If it compiles  it should run.

PR #78, opened 2023-05-06
Created an Eventsystem for time ticks.

Frozenenenen:
I created event channels for ingame seconds, minutes and hours. GameTime.cs now raises events on those channels.
The sun movement was deleted from GameTime.cs and is now in SunMover.cs

Sun movement every 10 ingame minutes was deleted for convinience reasons.

Additionally to the DateTime Event ScriptableObjects I created 2 eventsystems for int and float values that just exist as SO templates for now.


Assets/XR/Settings/OpenXR Package Settings.asset
JaXt0r, line 1: I know it's annoying, when Unity tells changes on files you didn't touch, but please try to not commit these files if you didn't want to change them. Don't know what could break in the future.

Packages/packages-lock.json
JaXt0r, line 1: I remember Tyroonis and Frank changed version on main or so. Maybe fetch their latest changes instead of changing it on your own?

Assets/GothicVR/Scripts/World/SunMover.cs
JaXt0r, line 1: I would recommend you put SunMover script onto Sun GameObject. As it's tightly coupled with it. And you don't need to set the GameObject as a property as its its own this.gameObject.

Assets/GothicVR/Scripts/World/SunMover.cs
```diff
@@ -0,0 +1,61 @@
...
+
+    public DateTimeGameEvent secondChangeEvent;
+    public DateTimeGameEvent minuteChangeEvent;
+    public DateTimeGameEvent hourChangeEvent;
+
+    private void OnEnable()
+    {
+
+    }
+    private void OnDisable()
```
JaXt0r, line 21: Thanks for thinking about the dynamic on/off feature.
Why not putting the OnEventRaised += ... method to the OnEnable() function as the disable is also on Disable()?

Assets/GothicVR/Scripts/World/SunMover.cs
```diff
@@ -0,0 +1,61 @@
...
+            case DebugSettings.SunMovementPerformance.EveryIngameHour:
+                hourChangeEvent.OnEventRaised += RotateSun;
+                break;
+            default:
+                break;
+        }
+    }
+    void RotateSun(DateTime time)
+    {
+        var xRotation = 270f + (15f * (time.Hour + (time.Minute / 60f) + (time.Second / 3600f)));
```
JaXt0r, line 58: I remember this line... :-D
Please re-add the extensive comment I created in my old version. Otherwise reviewing this line will be a pain in a month or later.

Assets/GothicVR/Scripts/World/SunMover.cs
```diff
@@ -0,0 +1,61 @@
...
+            case DebugSettings.SunMovementPerformance.EveryIngameSecond:
+                secondChangeEvent.OnEventRaised += RotateSun;
+                break;
+            case DebugSettings.SunMovementPerformance.EveryIngameMinute:
+                minuteChangeEvent.OnEventRaised += RotateSun;
+                break;
+            case DebugSettings.SunMovementPerformance.EveryIngameHour:
+                hourChangeEvent.OnEventRaised += RotateSun;
+                break;
+            default:
```
JaXt0r, line 52: I recommend throwing an exception or add a debug.error statement. Otherwise we won't get informed when there is a new setting and sun movement is broken based on this new "unrecognized" setting.

Assets/GothicVR/Scripts/SO Template/DateTimeGameEvent.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using UnityEngine;
+using UnityEngine.Events;
+
+[CreateAssetMenu(menuName = "Events/DateTime Event Channel")]
```
JaXt0r, line 5: Just checking: Is "channel" the right wording? Or would "Event"/"Callback"/etc. be a better name?

Assets/GothicVR/Scripts/World/GameTime.cs
```diff
@@ -11,8 +11,13 @@ public class GameTime : SingletonBehaviour<GameTime>
         private static DateTime MIN_TIME = new(1, 1, 1, 0, 0, 0);
         private static DateTime MAX_TIME = new(1, 1, 1, 23, 59, 59);
 
-        private DebugSettings.SunMovementPerformance sunPerformanceSetting;
+        public DateTimeGameEvent secondChangeChannel;
```
JaXt0r, line 14: Why don't we just use Unity's built in UnityAction event? What's the benefit of this additional abstraction/hop?

I used UnityEvents in this class as well. Worked fine:
https://github.com/GothicVRProject/GothicVR/blob/main/Assets/GothicVR/Scripts/Phoenix/Interface/Vm/VmGothicBridge.cs#L48

Assets/GothicVR/Scripts/World/GameTime.cs
```diff
@@ -21,16 +26,25 @@ public class GameTime : SingletonBehaviour<GameTime>
         private static readonly float ONE_INGAME_SECOND = 0.06944f;
         private DateTime time = new(1, 1, 1, 15, 0, 0);
 
-        public GameObject sun;
+        private void OnEnable()
+        {
+            //Subscribe to Events
+            secondChangeChannel.OnEventRaised += RaiseMinuteEvent;
```
JaXt0r, line 32: Please think about having this easier. It took me a while to understand why the secondEvent raises the MinuteEvent etc.

Why not having one method which is called every ingame second (you have it already) and then check there inside if a minute or hour is to be raised.

If there's value of this abstraction I don't understand: Cool. If not, let's think about the more direct approach as mentioned above.

Assets/GothicVR/Scripts/SO Template/DateTimeGameEvent.cs
```diff
@@ -0,0 +1,14 @@
...
+
+public class DateTimeGameEvent : ScriptableObject
+{
+    public UnityAction<DateTime> OnEventRaised;
```
JaXt0r, line 9: There's a more detailed comment below. But in short: Is this abstraction really needed with the additional Event classes? Why not leveraging Unity's events directly?

Assets/GothicVR/Scripts/World/SunMover.cs
JaXt0r, line 1: PS: I like you extracted sun movement to a separate Component instead of having it inside the GameTime Component itself.  👍

JaXt0r (changes requested):
I added a few complexity questions and naming ideas. Please have a look.
I also tested it and second / minute - sun movement works fine.

Packages/packages-lock.json
JucanAndreiDaniel: This wasn't pushed by us, but it doesn't affect anything afaik since it only has what packages are installed locally.

Frozenenenen: shouldnt we put it into the gitignore then?

Assets/GothicVR/Scripts/World/SunMover.cs
Frozenenenen: Done

Assets/GothicVR/Scripts/World/GameTime.cs
Frozenenenen: Altered to UnityEvents.

Assets/GothicVR/Scripts/World/GameTime.cs
Frozenenenen: Done

Assets/GothicVR/Scripts/World/SunMover.cs
Frozenenenen: I actually don't know what to check for.
Well it doesn't look like this anymore anyways.

JaXt0r: The idea is:
```csharp
switch()
    case ...:
        ...
        break;
    default:
        Debug.LogError($"SunMovement setting {sunPerformanceSetting} isn't handled right now. Therefore sun won't move.");
        break;
```

JaXt0r (changes requested):

Packages/packages-lock.json
JaXt0r: Not an expert, but I think this file should be Out on git. But should lily be changed if we actually change the manifest.json as well.
I think it's ok for now.

JaXt0r (approved):
Nice!

JucanAndreiDaniel (approved):
LGTM

PR #81, opened 2023-05-08
Feature/74 additional vob imports for chests ladders etc

JaXt0r:
More VOBs are now created.

e.g.:
* Benches - rotated right
* fireplace + chickensticks - rotated wrongly
* a lot of furniture in the PSI camp - roated right

I will take care of the rotation stuff later. But you can already use the additional VOBs.

PS: I updated all the shared libraries. Please use them on the main branch.


Tyroonis (changes requested):
The game crashes at startup.
Is it possible that libraries are missing?

JucanAndreiDaniel:
> The game crashes at startup. Is it possible that libraries are missing?

Fixed the issue until animations are done properly for Npcs, it was the bones that seem to have caused the crash

Tyroonis (approved):

Tyroonis:
Works on Pico4

PR #86, opened 2023-05-12
73 expand log script

Tyroonis:
First implementation of the new LogSystem.

Logs are written to the data folder on the glasses.
On PC in ...\AppData\LocalLow\GothicVRProject\GothicVR.

Currently everything is logged.
Log is overwritten at each startup, so that the memory does not run full.

For adjustments and log level from Config get I make a new issue.

@JucanAndreiDaniel  please see if it works on the quest


JucanAndreiDaniel (approved):
LGTM, works fine on Quest and also on PCVR build

JaXt0r:
> LGTM, works fine on Quest and also on PCVR build

Can you create the documentation change, please. You know best what to tell.

PR #88, opened 2023-05-12
5 npc routines improve performance and routing

Frozenenenen:
Routines now get called by the IngameMinute Tick Event.
Waypoints are now addtionally in a Dictionary. The List still exists parallel
Access in NPC Creator and Routine mostly changed to the Dictionary. (The if exist in the beginning still uses the list)
Addet a Debug Option to en/disable the NPC creation. Enabled most options in the debugsettings, what's displayed in the scene (I guess that doesnt mater)


Frozenenenen:
I couldnt Delete the waypoints lists without changing the waynet creation.

Assets/GothicVR/Scripts/Npc/Routine.cs
```diff
@@ -7,44 +7,65 @@
...
+            if (!SingletonBehaviour<DebugSettings>.GetOrCreate().EnableNpcRoutines)
+                return;
 
-            return routine;
+            getRoutine(time);
+            if (routine == null)
+                return;
+            getWaypoint();
+        }
+        void getRoutine(DateTime time)
```
JaXt0r, line 62: What you do is "setRoutine()" or "calculateRoutine()". But you don't "get" it back.

I think it's fine your calculate the routine time inside and store the current variable. But I would recommend renaming the function.

Assets/GothicVR/Scripts/Npc/Routine.cs
```diff
@@ -7,44 +7,65 @@
...
+        {
+            routine = routines.FirstOrDefault(item => (item.start <= time && time < item.stop));
+        }
+        void getWaypoint()
+        {
+            waypoint = PhoenixBridge.World.waypointsDict[routine.waypoint];
```
JaXt0r, line 68: It's also a "set/calculate" not a "get".

Assets/GothicVR/Scripts/Npc/Routine.cs
```diff
@@ -7,44 +7,65 @@
...
 namespace GVR.Npc
 {
     public class Routine : MonoBehaviour
     {
-        private static readonly int SPEED = 10;
-        private GameTime gameTime;
+        private const float SPEED = 1f;
+        private GameTime gameTime = new();
+        PxCs.Data.WayNet.PxWayPointData waypoint;
+        RoutineData routine;
```
JaXt0r, line 20: This is the cached "currently active routine", righ? I recommend you name it like this. "currentRoutine" or "activeRoutine". This eases reading in the future.

Assets/GothicVR/Scripts/Npc/Routine.cs
```diff
@@ -7,44 +7,65 @@
...
 
-        void Update()
+        private void Start()
         {
-            if (!SingletonBehaviour<DebugSettings>.GetOrCreate().EnableNpcRoutines)
-                return;
-
-            var routine = GetCurrentRoutine();
+            //Initialize first waypoint
+            DateTime time = new(1, 1, 1, 15, 0, 0);
```
JaXt0r, line 33: This datetime seems like a const. You could think of putting this as a const variable on top of the class.

Assets/GothicVR/Scripts/Creator/NpcCreator.cs
```diff
@@ -67,11 +68,10 @@ public static void Wld_InsertNpc(int npcInstance, string spawnpoint)
 
             if (PhoenixBridge.npcRoutines.TryGetValue(pxNpc.npcPtr, out List<RoutineData> routines))
```
JaXt0r, line 69: Can we get rid of this additional temp variable?

I mean we already have the data. We just need to look it up:
```csharp
newNpc.GetComponent<Routine>().routines.First();
```

Assets/GothicVR/Scripts/Phoenix/Interface/WorldBridge.cs
```diff
@@ -28,6 +29,11 @@ public static WorldData LoadWorld(IntPtr vdfsPtr, string worldName)
             var materials = PxMesh.GetMaterials(worldMeshPtr);
 
             var waypoints = PxWorld.GetWayPoints(worldPtr);
+            var waypointsDict = new Dictionary<string, PxWayPointData>();
```
JaXt0r, line 32: Cool performance optimization!

Assets/GothicVR/Scripts/Phoenix/Interface/PhoenixBridge.cs
```diff
@@ -18,5 +19,6 @@ public static class PhoenixBridge
 
         // FIXME Find a better place for the NPC routines. E.g. on the NPCs itself? But we e.g. need to have a static NPCObject List to do so.
         public static Dictionary<IntPtr, List<RoutineData>> npcRoutines = new();
+        public static Dictionary<IntPtr, Dictionary<DateTime, RoutineData>> npcRoutinesDict = new();
```
JaXt0r, line 22: Do we use this? Or: Can we get rid of it? The data is already inside npcObj.GetComponent<Routine>()

JaXt0r (changes requested):

Assets/GothicVR/Scripts/Npc/Routine.cs
Frozenenenen: It is just needed to intialize the startposition. so it gets deleted when Start() is finished anyway.

Assets/GothicVR/Scripts/Phoenix/Interface/PhoenixBridge.cs
Frozenenenen: yep, obsolete - deleted

Assets/GothicVR/Scripts/Npc/Routine.cs
Frozenenenen: I tried. DateTime can't be made const.

JaXt0r: Yes. It's about "magic numbers": https://en.wikipedia.org/wiki/Magic_number_(programming)

The idea is: If there is a number which has a special meaning, give it a useful name.
To give a name, you need to define a property.

I think of e.g.
```private const DateTime _MinimumDateTime_ = ...```
or
```private const DateTime _StartDateTimeBecause... = ..._```

File: Assets/GothicVR/Scripts/Creator/NpcCreator.cs (hunk 5 above)
Comment (Frozenenenen) [reply Assets/GothicVR/Scripts/Creator/NpcCreator.cs:69]: I don't get, what you are up to with this one

File: Assets/GothicVR/Scripts/Creator/NpcCreator.cs [hunk 8]
```diff
@@ -130,8 +130,11 @@ private static void Mdl_SetVisualBody(VmGothicBridge.Mdl_SetVisualBodyData data)
             var npc = lookupCache.npcCache[symbolIndex];
             var mdh = npc.GetComponent<Properties>().mdh;
             var mdm = assetCache.TryGetMdm(data.body);
-            
-            // SingletonBehaviour<MeshCreator>.GetOrCreate().Create(name, mdm, mdh, default, default, npc);
+
+            if (SingletonBehaviour<DebugSettings>.GetOrCreate().EnableNpc)
+                SingletonBehaviour<MeshCreator>.GetOrCreate().Create(name, mdm, mdh, default, default, npc);
+
+            meshCreator.Create(name, mdm, mdh, default, default, npc);
```
JaXt0r, line 137: Please check. It seems you added the same statement twice. Feel free to use this if it makes sense:

```suggestion

            if (SingletonBehaviour<DebugSettings>.GetOrCreate().EnableNpc)
                meshCreator.Create(name, mdm, mdh, default, default, npc);
```

Assets/GothicVR/Scripts/Creator/NpcCreator.cs
```diff
@@ -65,12 +67,8 @@ public static void Wld_InsertNpc(int npcInstance, string spawnpoint)
 
             newNpc.GetComponent<Properties>().npc = pxNpc;
 
-            if (PhoenixBridge.npcRoutines.TryGetValue(pxNpc.npcPtr, out List<RoutineData> routines))
-            {
-                initialSpawnpoint = PhoenixBridge.World.waypoints
-                    .FirstOrDefault(item => item.name.ToLower() == routines.First().waypoint.ToLower());
-                newNpc.GetComponent<Routine>().routines = routines;
-            }
+            var initialSpawnpointName = newNpc.GetComponent<Routine>().routines.First().waypoint;
```
JaXt0r, line 70: Is it always delivering a value? Or can .routines().First() be empty?
If so, then please add an if statement again.

Assets/GothicVR/Scripts/Npc/Routine.cs
```diff
@@ -7,44 +7,65 @@
...
         }
 
-        void Update()
+        private void Start()
         {
-            if (!SingletonBehaviour<DebugSettings>.GetOrCreate().EnableNpcRoutines)
-                return;
-
-            var routine = GetCurrentRoutine();
+            //Initialize first waypoint
```
JaXt0r, line 32: Good compromise. Hehe. :-)

JaXt0r (changes requested):
Nearly done. Two remaining bugs, if I see it correctly.

Assets/GothicVR/Scripts/Creator/NpcCreator.cs
Frozenenenen: my code actually doesnt look like that xD

Assets/GothicVR/Scripts/Creator/NpcCreator.cs
Frozenenenen: I changed it, but I think its superflous anyways, since at the top of the function there's already another intialSpawnpoint calculation. It works differently though and I dont know where the input comes from, so I cant be completely sure.

Frozenenenen:
I just saw that there are unidentified bugs and I won't be able to fix that now. I'll have to look tomorrow

PR #90, opened 2023-05-14
For all meshes we now use a cutout shader by default.

JaXt0r:
Please check it briefly:
* Wall on old camp
* plants
* and trees are now rendered cutout

image

### DIFF
--- a/Assets/GothicVR/Scripts/Creator/MeshCreator.cs
+++ b/Assets/GothicVR/Scripts/Creator/MeshCreator.cs
@@ -17,7 +17,7 @@ namespace GVR.Creator
         private static AssetCache assetCache;
 
         // DEBUG - we can change special mesh entries (trees, walls) based on flags later. But for now we can live with the nature cutout shader.
-        private const string DEFAULT_SHADER = "Nature/Tree Creator Leaves Fast";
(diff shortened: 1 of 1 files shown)


Assets/GothicVR/Scripts/Creator/MeshCreator.cs
```diff
@@ -16,6 +16,9 @@ public class MeshCreator : SingletonBehaviour<MeshCreator>
     {
         private static AssetCache assetCache;
 
+        // DEBUG - we can change special mesh entries (trees, walls) based on flags later. But for now we can live with the nature cutout shader.
+        private const string DEFAULT_SHADER = "Nature/Tree Creator Leaves Fast";
```
JucanAndreiDaniel, line 20: This shader seems to have some issues with the shadows or ghosting effect
image

If we remove the fast option it seems to solve the issue but not entirely as i've noticed it a little bit when roaming around the scene
image

```suggestion
        private const string DEFAULT_SHADER = "Nature/Tree Creator Leaves";
```

JucanAndreiDaniel (changes requested):

JaXt0r:
The example suits me. I think it's better than before and fine for now, right?

JucanAndreiDaniel (approved):
Yep, it's good.

PR #92, opened 2023-05-15
fix docs entry - remove unused

Tyroonis:
Github Actions did not cope with regular expression.

regular expression removed.
If we should adjust the name of the docs folder, the name must be adjusted here as well


Tyroonis:
Testet on Fork Projekt

Issue #63, opened 2023-04-29
Create list of Gothic Daedalus calls for docs


Thogsit:
The list should contain all calls that we have to reimplement in Unity. It should be used to document our current progress with implementing them and to have a single place to add notes and ideas regarding their implementation. Also, it can give an overview how well these are already optimized and which are good targets for refactoring/improvements.

Firstly, meet with @JaXt0r to discuss this further.

EDIT: List of all external Daedalus functions can be found [here](https://www.worldofgothic.de/modifikation/index.php?go=g2functions)

PR #94, opened 2023-05-15
docs: Add Daedalus external functions page

Thogsit:
Relates to #63, see Discord topic "Daedalus - Implement external functions for Unity"


JaXt0r (changes requested):
Tables look fine.

A few comments on formatting:
* I would recommend against TODO and DONE tables. It clutters the view.
* How about marking the status columns with colors? red/yellow/green would be sufficient and fast to see. Is anything of these working? https://stackoverflow.com/questions/11509830/how-to-add-color-to-githubs-readme-md-file
* Can you add a Table of Contents (TOC) where we can click on the entries and will be routed to the headline on the page (e.g. World)?

Romeeey:
Looks also good to me.

Same as @JaXt0r I recommend to not split it in todo and done. The status is enough to see the progress.

I would try to keep the page as English as possible, since we also have international devs.

There are really a lot of methods. We have to do some priorisation. It might help to classify some methods as urgent (they will still be 20-40) and some as not necessary or not important. The more I think about it, I want an Excel or some table function to filter.

JaXt0r:
English: The method descriptions are from Piranha Bytes itself. We would also need to do google translation on it. 😋

Prioritization: At least from my end, I don't know what's of priority. We need to know more about it before suggesting it in general. Or...
My idea would be: When we call one Daedalus functions, it will trigger more. If one of these called functions isn't handled, we get a warning in unity and can handle it then. I.e. prioritization is handled for us. 🫣 Example: Spawning NPCs in old Camp triggers ~10 functions which aren't handled right now. This can be our priority.

Thogsit:
In respect to the comments, I changed the following:
- Removed TODO/DONE sections
- Colored all "In Progress" entries red; LaTeX formatting worked without external services and seems to be good enough
- Added a Table of Contents which links to every section

The function descriptions are copy/pasted from the Gothic code with the intermediary source WoG and as such are in the original German language.

Personally, I'd like to prioritise the functions; but I think, at the moment everyone who works on the functions is active enough in the Discord to be informed what functions are currently blocking the current features. If someone encounters a blocking function and can't or doesn't want to work on it, I'd recommend just pinging in the Discord and asking whether someone's willing to work on it.
If we get new participants who primarily retrieve their information from this docs page, I'd rethink whether we add explicit priorities, but for now I would start with the page as is and see if we feel that something's missing in the next days/weeks.

Tyroonis (approved):

JaXt0r (approved):
I'd have never expected latex formatting is working in GitHub markdown 🤣

Two minor things:
* The column status and complexity is broken for Wld_InsertNpc
* Could you name the possible status fields inside Initial description?
* and recolour the status fields? I think of Green=Done, Orange=In Progress, Red=Todo (Feel free to suggest additionally)

I already approve as the suggestions are code candies. 😋

JaXt0r:
Hi Thogs. Any update on this?

Tyroonis (approved):

Tyroonis:
Does it make sense to write who is working on methods that are in progress?

JaXt0r:
I don't think people will update it tbh. I think Last time we agreed putting the ticket ID in here if useful.

But to be honest: If someone's working on it (in progress) and we want to work there too, we can just ask the team who owns it.

Thogsit:
> Hi Thogs. Any update on this?

I'm back today after being away for a few days.

Adjusted the docs according to your suggestions; I'll merge now, so it can be used.

PR #100, opened 2023-05-24
Unsubscribe throws errors as GameSettings object is disposed already.…

JaXt0r:
… Now we check unsubscribe values with TryGet()

Small bugfix to get rid of the errors.

### DIFF
--- a/Assets/GothicVR/Scripts/Npc/RoutineManager.cs
+++ b/Assets/GothicVR/Scripts/Npc/RoutineManager.cs
@@ -50,14 +50,16 @@ public class RoutineManager : SingletonBehaviour<RoutineManager>
 
     public void Unsubscribe(Routine routineInstance, List<RoutineData> routines)
     {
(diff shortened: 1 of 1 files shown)


Frozenenenen:
I don't get why you deleted the debug options. The Debug-option isn't really needed, but isn't it clearer that there's nothing happening, with the flag unset?

JaXt0r:
You're right. I didn't add a note.

It's because of a race condition: When you stop Unity play mode, the MonoBehaviours get wiped from memory. As we do Unregister() during this time, the DebugSettings MonoB. is already gone and tries itself to re-create which causes Error logs in Unity.

The change is mitigating this errors by:
* As we check on Register() if the setting is active, we don't need to have the check during Unregister()
* I added TryGetValue() so that the system will recognize empty routines Dict and just ignores it if the setting was off before.

Frozenenenen (approved):
LGTM

PR #103, opened 2023-06-08
Feature/101 chest implement loot

JaXt0r:
Hi folks. For the chest loot feature, I put quite some changes into the vob loading logic. Which is also reflected on PxCs already.

Please briefly review and accept if ok. Then I can put these commits into our other branches like the sound feature, where it's needed.


JucanAndreiDaniel (approved):
LGTM

PR #108, opened 2023-06-10
Enhancement/104 change rendering to universal render pipeline

### LINKED ISSUE #104: Change Rendering to Universal Render Pipeline
Current GVR uses the old render pipeline. Let's switch to the latest one to:
1. Have performance improvements (?)
2. Get GPU Profiler
3. Mobile VR won't crash with a lot of items/bones any longer (possibly)

Tyroonis:
This is an initial implementation of the URP.
Has already brought a lot. But must be optimized later.

- add Universal Render Pipeline
- switch to URP Shader
- add material so that shader builds it into the build

Known problems:
- Invisible walls visible again
- Trees and bushes texture also does not fit
Will be worked on in separate issue (#107)

close #104

### DIFF
--- a/Assets/GothicVR/Resources/Prefabs/NPC.prefab
+++ b/Assets/GothicVR/Resources/Prefabs/NPC.prefab
@@ -65,7 +65,7 @@ MeshRenderer:
   m_RenderingLayerMask: 1
   m_RendererPriority: 0
   m_Materials:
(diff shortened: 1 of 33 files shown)


Tyroonis:
@Romeeey or @JucanAndreiDaniel 
Please check Pipeline Quest Build on your Device.

JaXt0r (approved):

Issue #6, opened 2023-03-29
Worldmusic - Play gothic music track


JaXt0r:
We want to have Gothic music being played when walking through the barrier.
* [x] Check within [phoenix](https://github.com/lmichaelis/phoenix/tree/30179d81a678f8a36fbaf3306a443f779520a0a0) and [OpenGothic](https://github.com/Try/OpenGothic) how it was loaded there
* [x] Build a bridging functionality to load the track from C++ into C#
* [x] Store the binary wav stream in a (e.g.) Unity AudioClip and play
* [x] Create VOB with triggers to play music in a dynamic way
* [x] Create object to handle and music state and change

Boundary:
This ticket does not include dynamic track change. A first atmospheric and iconic gothic sound would be awesome. ♥

Hint:
Gothic leverages music as DirectMusic format. We need to check if it's
1. possible to use within Unity,
2. buy an asset to have it parsed, or
3. rebuild ffmpeg/OpenGothic's implementation: https://github.com/Try/OpenGothic/tree/master/game/dmusic

JucanAndreiDaniel:
http://www.vgmpf.com/Wiki/index.php?title=SGT
http://www.vgmpf.com/Wiki/index.php?title=STY
http://www.vgmpf.com/Wiki/index.php?title=DLS
Some documentation which shows the RIFF Tree Structure of the music formats used in Gothic

https://documentation.help/DirectMusic/  - docs for everything related to DirectMusic (inferred from what I skimmed through)

JucanAndreiDaniel:
Checked the original tasks and added some more to reflect current progress

PR #127, opened 2023-06-20
Feature/audio implementation

JucanAndreiDaniel:
Implement 
1. sound effects which are created in game
2. dialogue WAV decoding to be used in game
3. Initial music integration (will come fully fleshed in another PR)

PR #129, opened 2023-06-25
Music implementation

JucanAndreiDaniel:
Continue music implementation from #127 

Add Music switching based on zones


JucanAndreiDaniel:
should close #6

Assets/GothicVR/Scripts/Importer/PhoenixImporter.cs
```diff
@@ -116,9 +117,16 @@ private void LoadSfxVM(string G1Dir)
...
         {
-            SingletonBehaviour<MusicTest>.GetOrCreate().Create(G1Dir);
+            var fullPath = Path.GetFullPath(Path.Join(G1Dir, "/_work/DATA/scripts/_compiled/MUSIC.DAT"));
+            var vmPtr = VmGothicBridge.LoadVm(fullPath);
+            PhoenixBridge.VmMusicPtr = vmPtr;
+        }
+
+        private void LoadMusic(string G1Dir)
+        {
+            SingletonBehaviour<MusicCreator>.GetOrCreate().Create(G1Dir);
```
JaXt0r, line 129: What is it doing? I see you have .Create() method in here but don't return anything. To be aligned with the other Creator.cs classes, please return the value you loaded and store it somewhere.
example: _MeshCreator_ or _VobCreator_

Assets/GothicVR/Scripts/Creator/MusicCreator.cs
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        private GameObject CreateSoundObject()
+        {
+            return new GameObject("Background Music");
+        }
+
+        private AudioClip CreateAudioClip()
+        {
+            return AudioClip.Create("Music", bufferSize * 4, 2, 44100, true, PrepareData);
```
JaXt0r, line 82: Please add a comment what these magic numbers are.

Assets/GothicVR/Scripts/Creator/MusicCreator.cs
```diff
@@ -0,0 +1,216 @@
...
+
+        private void AddMusicPath(string fullPath, string path)
+        {
+            fullPath = Path.Combine(fullPath, path);
+            DMDirectMusic.DMusicAddPath(directmusic, fullPath);
+        }
+
+        private GameObject CreateSoundObject()
+        {
+            return new GameObject("Background Music");
```
JaXt0r, line 77: Will there only be one _Background Music_ GameObject or multiple? If we have multiple, think of naming it with the audio file name etc. Like we do with meshes. They all have a special name.

Assets/GothicVR/Scripts/Phoenix/Util/CheckIfAudible.cs
```diff
@@ -0,0 +1,42 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+namespace GVR.Phoenix.Util
+{
+    public class CheckIfAudible : MonoBehaviour
```
JaXt0r, line 6: Please: rename it so that it reflects the use case better. Like: _InaudibleSoundDisabler_

Assets/GothicVR/Scripts/Phoenix/Util/CheckIfAudible.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        AudioListener audioListener;
+        AudioSource audioSource;
+        float distanceFromPlayer;
```
JaXt0r, line 10: Private or public? I think private is sufficient here, right?

Assets/GothicVR/Scripts/Phoenix/Util/CheckIfAudible.cs
```diff
@@ -0,0 +1,42 @@
...
+        void Start()
+        {
+            // Finds the Audio Listener and the Audio Source on the object
+            audioListener = Camera.main.GetComponent<AudioListener>();
+            audioSource = gameObject.GetComponent<AudioSource>();
+        }
```
JaXt0r, line 16: Please add some empty line after methods.

Assets/GothicVR/Scripts/Phoenix/Util/CheckIfAudible.cs
```diff
@@ -0,0 +1,42 @@
...
+        void Update()
```
JaXt0r, line 17: Ok. I see your point. I'm just curious if this will cause performance issues when we do checks for every sound in the world every frame. But maybe this is nothing to be solved here...
My ideas for the future (just FYI):
1. We spawn world objects only if we reached a certain area (like when in OC, we have OC vobs only and the rest is disabled/removed)
2. You could also set a Coroutine with a condition WaitUntil() then it's only triggered if condion is met. (But in the end the mathematics will be the same. Just at another spot 🤔 )

Do you see a problem here?

Assets/GothicVR/Scripts/Phoenix/Util/MusicColliderHandler.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        private static List<string> musicZones = new List<string>();
+
+        void OnTriggerEnter(Collider other)
+        {
+            if (!other.CompareTag("Player")) return;
+
+            musicZones.Add(gameObject.name);
+
+            SingletonBehaviour<MusicCreator>.GetOrCreate().SetMusic(gameObject.name, MusicCreator.Tags.Std);
```
JaXt0r, line 18: Is it creating, loading, or just setting a value?
Can you be more precise, please. As the creators were for creation only before.

ProjectSettings/AudioManager.asset
```diff
@@ -10,8 +10,9 @@ AudioManager:
   Default Speaker Mode: 2
   m_SampleRate: 0
   m_DSPBufferSize: 1024
-  m_VirtualVoiceCount: 512
+  m_VirtualVoiceCount: 1024
   m_RealVoiceCount: 32
+  m_EnableOutputSuspension: 1
```
JaXt0r, line 15: What does this do? Will it stop playing Audio when inaudible? Is this intended?

JaXt0r (changes requested):

ProjectSettings/AudioManager.asset
JucanAndreiDaniel: > Enable Output Suspension (editor only)", "When enabled automatically suspends audio output after detecting that the output has been silent for a long duration (editor only).

taken from https://github.com/Unity-Technologies/UnityCsReference/blob/master/Editor/Mono/Inspector/AudioManagerInspector.cs#L29

it was added by Unity when i modified the m_VirtualVoiceCount in the editor

can be removed

Assets/GothicVR/Scripts/Phoenix/Util/MusicColliderHandler.cs
JucanAndreiDaniel: Since we are using the Singleton pattern, we can use the object created by [PhoenixImporter.cs](https://github.com/GothicVRProject/GothicVR/blob/6ae529fffc7df0b16c5144f3e56f8628c57f2d9d/Assets/GothicVR/Scripts/Importer/PhoenixImporter.cs#L129) everywhere, since it's supposed to only be one instance, to mix and update the music.

Assets/GothicVR/Scripts/Phoenix/Util/CheckIfAudible.cs
JucanAndreiDaniel: I don't believe the current approach is that computationally expensive. We could profile this implementation to see if it affects the performance on Quest and Pico, but from my testing it doesnt seem to be an issue

Assets/GothicVR/Scripts/Creator/MusicCreator.cs
JucanAndreiDaniel: There should be only one background music object per scene.
I'm thinking of removing this function since it doesn't serve a clearer purpose, could be same for CreateAudioClip

Assets/GothicVR/Scripts/Importer/PhoenixImporter.cs
JucanAndreiDaniel: Moved music initialization logic from Create to Start to so we do not need this function anymore. Music will automatically be created and checking for the debug flag to play the music or not

Assets/GothicVR/Scripts/Creator/MusicCreator.cs
JucanAndreiDaniel: Done

Assets/GothicVR/Scripts/Creator/MusicCreator.cs
JucanAndreiDaniel: Moved to a dedicated GameObject for the background music

Assets/GothicVR/Scripts/Creator/SoundCreator.cs
```diff
@@ -55,6 +55,24 @@ public GameObject Create(PxVobSoundData vobSound, GameObject parent = null)
...
+        {
+            var soundObject = new GameObject(vobSound.vobName);
+            soundObject.SetParent(parent);
+
+            var soundObjectCollider = soundObject.AddComponent<BoxCollider>();
+
+            var min = vobSound.boundingBox.min.ToUnityVector();
+            var max = vobSound.boundingBox.max.ToUnityVector();
+
+            soundObject.transform.position = (min + max) / 2f;
```
JaXt0r, line 68: Why is it divided by 2?

JucanAndreiDaniel: to find the center of the bounding box as that is needed to properly place the gameobject

PR #133, opened 2023-07-02
Bugfix/music crash when no musicvm

JucanAndreiDaniel:
When creating the music component we might reference to a nullpointer of MusicVM and crash Unity so we're moving the initialization from Start to Create
Also initialize music back in PhoenixImporter.cs


JaXt0r (approved):
LGTM.

PR #137, opened 2023-07-05
Feature/unloadable vobs fix

JaXt0r:
What changed:
* Freepoints are now visible if activated on DebugSettings (round balls)
* Decals are now imported
* Shader moved from Unlit to URP/Unlit - Please check mobile performance

Can be tested now, but merged after https://github.com/GothicKit/phoenix-csharp-interface/pull/17 only.

---

# Checklist

## Scene

# Testing
* [ ] Tested with PCVR and one mobile VR device (Pico/Quest)

## Dependencies

### DIFF
--- a/Assets/GothicVR/Scripts/Creator/MeshCreator.cs
+++ b/Assets/GothicVR/Scripts/Creator/MeshCreator.cs
@@ -17,7 +17,9 @@ namespace GVR.Creator
     {
         private AssetCache assetCache;
 
-        private const string DEFAULT_SHADER = "Unlit/Transparent Cutout";
+        // Decals work only on URP shaders. We therefore temporarily change everything to this
+        // until we know how to change specifics to the cutout only. (e.g. bushes)
+        private const string defaultShader = "Universal Render Pipeline/Unlit"; // "Unlit/Transparent Cutout";
 
         private void Start()
         {
@@ -168,7 +170,7 @@ namespace GVR.Creator
 
         private void PrepareMeshRenderer(Renderer rend, WorldData.SubMeshData subMesh)
         {
-            var standardShader = Shader.Find(DEFAULT_SHADER);
+            var standardShader = Shader.Find(defaultShader);
             var material = new Material(standardShader);
             var bMaterial = subMesh.material;
 
@@ -202,7 +204,7 @@ namespace GVR.Creator
 
             foreach (var subMesh in mrmData.subMeshes)
             {
-                var standardShader = Shader.Find(DEFAULT_SHADER);
+                var standardShader = Shader.Find(defaultShader);
                 var material = new Material(standardShader);
                 var materialData = subMesh.material;
 
--- a/Assets/GothicVR/Scripts/Creator/MeshCreator.cs
+++ b/Assets/GothicVR/Scripts/Creator/MeshCreator.cs
@@ -170,10 +170,9 @@ namespace GVR.Creator
 
         private void PrepareMeshRenderer(Renderer rend, WorldData.SubMeshData subMesh)
         {
-            var standardShader = Shader.Find(defaultShader);
-            var material = new Material(standardShader);
+            var material = GetEmptyMaterial();
             var bMaterial = subMesh.material;
-
+            
             rend.material = material;
 
             // No texture to add.
@@ -204,8 +203,7 @@ namespace GVR.Creator
 
             foreach (var subMesh in mrmData.subMeshes)
             {
-                var standardShader = Shader.Find(defaultShader);
-                var material = new Material(standardShader);
+                var material = GetEmptyMaterial();
                 var materialData = subMesh.material;
 
                 rend.material = material;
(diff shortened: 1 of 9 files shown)


Assets/GothicVR/Scripts/Creator/VobCreator.cs
```diff
@@ -164,11 +196,76 @@ private void CreateZoneMusic(PxVobZoneMusicData vob)
...
+            SetPosAndRot(spot, vob.position, vob.rotation!.Value);
+        }
+
         private GameObject CreateItemMesh(PxVobItemData vob, PxVmItemData item, GameObject go)
         {
             var mrm = assetCache.TryGetMrm(item.visual);
             return meshCreator.Create(item.visual, mrm, vob.position.ToUnityVector(), vob.rotation!.Value, true, parentGos[vob.type], go);
         }
+
+        private void CreateDecal(PxVobData vob)
```
JucanAndreiDaniel, line 235: This function should be moved to MeshCreator

Assets/GothicVR/Scripts/Creator/VobCreator.cs
```diff
@@ -164,11 +196,76 @@ private void CreateZoneMusic(PxVobZoneMusicData vob)
             soundCreator.Create(vob, parentGos[vob.type]);
         }
 
+        /// <summary>
+        /// Basically a free point where NPCs can do something like sitting on a bench etc.
+        /// @see for more information: https://ataulien.github.io/Inside-Gothic/objects/spot/
+        /// </summary>
+        private void CreateSpot(PxVobData vob)
```
JucanAndreiDaniel, line 203: Add Debug Option to enable/disable the spots from rendering

Assets/GothicVR/Scripts/Creator/VobCreator.cs
```diff
@@ -4,22 +4,28 @@
...
+    using UnityEditor;
+#endif
+
+
 namespace GVR.Creator
 {
 	public class VobCreator : SingletonBehaviour<VobCreator>
-    {   
+    {
+        private const float decalOpacity = 0.5f;
```
JucanAndreiDaniel, line 27: decals should be 1f most of the time, exception should be maybe dirt on castle walls

Assets/GothicVR/Scripts/Creator/MeshCreator.cs
```diff
@@ -471,5 +471,16 @@ private static void CreateBonesData(GameObject root, SkinnedMeshRenderer rendere
             renderer.sharedMesh.bindposes = bindPoses;
             renderer.bones = bones;
         }
+        
+        private Material GetEmptyMaterial()
+        {
+            var standardShader = Shader.Find(defaultShader);
```
JucanAndreiDaniel, line 477: Check if defaultShader is loaded correctly as right now the shader cannot be found on builds and returns null

JucanAndreiDaniel (changes requested):
Mostly small changes requested :D

Assets/GothicVR/Scripts/Creator/MeshCreator.cs
JaXt0r: Found a way to reference it: https://stackoverflow.com/questions/36909567/textures-loads-in-editor-but-not-in-standalone-appears-pink

Assets/GothicVR/Scripts/Creator/VobCreator.cs
JaXt0r: I would disagree. We discussed within Discord and mentioned 0.5f-0.75f
At least the walls look fine with this value and I would not want to have it "adjusted" for specific decals as of now.

Suggestion: Let's stick with a value and optimize it once we have lightning done right.

Assets/GothicVR/Scripts/Creator/VobCreator.cs
JaXt0r: It's there. Just look into the method. It even adds labels if we're in Editor mode.

JaXt0r: Added comment for better visibility.

JaXt0r:
@JucanAndreiDaniel  Commented and solved. Please review again.

JucanAndreiDaniel (changes requested):
Maybe Universal Render Pipeline/Unlit is not the best solution for now, we could only leave the submeshes to use Universal Render Pipeline/Unlit and the VOB to use Unlit/Transparent Cutout

Assets/GothicVR/Scripts/Creator/MeshCreator.cs
```diff
@@ -471,5 +509,16 @@ private static void CreateBonesData(GameObject root, SkinnedMeshRenderer rendere
...
+            var material = new Material(standardShader);
+            
+            // Enable clipping of alpha values.
+            material.EnableKeyword("_ALPHATEST_ON");
```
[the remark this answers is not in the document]
JucanAndreiDaniel: My bad, the issue was actually generated by not adding the decal shader which stops unity from loading the rest of the assets.

JucanAndreiDaniel (approved):
LGTM now, tested standalone quest and pcvr

Issue #138, opened 2023-07-08
VOb creation error stops creating other VObs


JucanAndreiDaniel:
If by some chance when creating a VOb we get an error, the rest of the objects that can be created are not processed anymore.

PR #141, opened 2023-07-09
Bugfix/vob error catch

JucanAndreiDaniel:
Added some checks and logs for gameobjects which have a chance to throw exceptions and make the world not load properly.

## Scene

# Testing

### DIFF
--- a/Assets/GothicVR/Scripts/Caches/AssetCache.cs
+++ b/Assets/GothicVR/Scripts/Caches/AssetCache.cs
@@ -161,18 +161,7 @@ namespace GVR.Caches
             if (soundCache.TryGetValue(preparedKey, out PxSoundData<float> data))
                 return data;
             
-            PxSoundData<float> wavFile = new PxSoundData<float>();
-
-            try
-            {
-                wavFile = PxSound.GetSoundArrayFromVDF<float>(PhoenixBridge.VdfsPtr, $"{GetPreparedKey(key)}.wav");
-            }
-            catch
-            {
-                Debug.LogError("AccessViolationException: Sound not found.");
-                return null;
-            }
-            
+            var wavFile = PxSound.GetSoundArrayFromVDF<float>(PhoenixBridge.VdfsPtr, $"{GetPreparedKey(key)}.wav");
             soundCache[preparedKey] = wavFile;
 
             return wavFile;
--- a/Assets/GothicVR/Scripts/Caches/AssetCache.cs
+++ b/Assets/GothicVR/Scripts/Caches/AssetCache.cs
@@ -161,7 +161,17 @@ namespace GVR.Caches
             if (soundCache.TryGetValue(preparedKey, out PxSoundData<float> data))
                 return data;
             
-            var wavFile = PxSound.GetSoundArrayFromVDF<float>(PhoenixBridge.VdfsPtr, $"{GetPreparedKey(key)}.wav");
+            PxSoundData<float> wavFile = new PxSoundData<float>();
+
+            try
+            {
+                wavFile = PxSound.GetSoundArrayFromVDF<float>(PhoenixBridge.VdfsPtr, $"{GetPreparedKey(key)}.wav");
+            }
+            catch
+            {
+                Debug.LogError("AccessViolationException: Sound not found.");
+                return null;
+            }
             
             soundCache[preparedKey] = wavFile;
 
--- a/Assets/GothicVR/Scripts/Creator/MeshCreator.cs
+++ b/Assets/GothicVR/Scripts/Creator/MeshCreator.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using GVR.Caches;
@@ -221,7 +221,6 @@ namespace GVR.Creator
 
             // No texture to add.
             if (bMaterial.texture == "")
-
             {
                 Debug.LogWarning("No texture was set for: " + bMaterial.name);
                 return;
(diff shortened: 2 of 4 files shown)


JucanAndreiDaniel:
Solves #138

Assets/GothicVR/Scripts/Caches/AssetCache.cs
```diff
@@ -161,7 +161,17 @@ public PxSoundData<float> TryGetSound(string key)
             if (soundCache.TryGetValue(preparedKey, out PxSoundData<float> data))
                 return data;
             
-            var wavFile = PxSound.GetSoundArrayFromVDF<float>(PhoenixBridge.VdfsPtr, $"{GetPreparedKey(key)}.wav");
+            PxSoundData<float> wavFile = new PxSoundData<float>();
```
JaXt0r, line 164: It's not needed to initialize the variable. This is sufficient:

```suggestion
            PxSoundData<float> wavFile;
```

Assets/GothicVR/Scripts/Caches/AssetCache.cs
```diff
@@ -161,7 +161,17 @@ public PxSoundData<float> TryGetSound(string key)
...
+
+            try
+            {
+                wavFile = PxSound.GetSoundArrayFromVDF<float>(PhoenixBridge.VdfsPtr, $"{GetPreparedKey(key)}.wav");
```
JaXt0r, line 168: Do we need a try/catch? Why not returning a null? I feel PxCs should take care of it. Other phoenix-shared calls always return null if neccessary. So no "error" handling needed.

Assets/GothicVR/Scripts/Creator/MeshCreator.cs
```diff
@@ -139,12 +139,18 @@ public GameObject Create(string objectName, PxMultiResolutionMeshData mrm, Vecto
             rootGo.SetParent(parent);
             SetPosAndRot(rootGo, position, rotation);
 
+            if (mrm == null)
+            {
+                Debug.LogError("No mesh data was found for: " + objectName);
+                return rootGo;
```
JaXt0r, line 145: If we have no mesh I would recommend we return null instead of the empty GameObject. This way we can easily handle in the calling function what to do.

PS: If you want to return null, please take care, that you did Destroy(rootGo) otherwise we have garbage in the scene.

Assets/GothicVR/Scripts/Creator/MeshCreator.cs
```diff
@@ -210,17 +216,25 @@ private void PrepareMeshRenderer(Renderer rend, WorldData.SubMeshData subMesh)
...
+                Debug.LogWarning("No texture was set for: " + bMaterial.name);
                 return;
+            }
 
             var texture = assetCache.TryGetTexture(bMaterial.texture);
 
             if (null == texture)
-                throw new Exception("Couldn't get texture from name: " + bMaterial.texture);
+                // throw new Exception("Couldn't get texture from name: " + bMaterial.texture);
+                if (bMaterial.texture.EndsWith(".TGA"))
```
JaXt0r, line 234: That's cool check.

Assets/GothicVR/Scripts/Creator/SoundCreator.cs
```diff
@@ -114,6 +114,13 @@ private AudioSource CreateAndAddAudioSource(GameObject soundObject, string sound
             else
             {
                 var sfxData = assetCache.TryGetSfxData(soundData.soundName);
+
+                if (sfxData == null)
+                {
+                    Debug.LogError($"No sfx data returned for {soundData.soundName}");
+                    return null;
```
JaXt0r, line 121: Ah! One more thing: Please don't return null! Please set the cache key. Otherwise it will try to load it again and again.

Assets/GothicVR/Scripts/Caches/AssetCache.cs
```diff
@@ -161,7 +161,17 @@ public PxSoundData<float> TryGetSound(string key)
...
+            }
+            catch
+            {
+                Debug.LogError("AccessViolationException: Sound not found.");
+                return null;
```
JaXt0r, line 173: Ah! One more thing: Please don't return null! Please set the cache key. Otherwise it will try to load it again and again.

Assets/GothicVR/Scripts/Creator/VobCreator.cs
```diff
@@ -133,7 +133,16 @@ private void CreateItem(PxVobItemData vob)
 
             // e.g. ItMiCello is commented out on misc.d file.
             if (item == null)
+            {
+                Debug.LogError($"Item {itemName} not found.");
+                return;
+            }
+
+            if (item.visual.EndsWith(".mms"))
```
JaXt0r, line 141:
```suggestion
            if (item.visual.ToLower().EndsWith(".mms"))
```

Assets/GothicVR/Scripts/Creator/VobCreator.cs
```diff
@@ -133,7 +133,16 @@ private void CreateItem(PxVobItemData vob)
...
+            {
+                Debug.LogError($"Item {item.visual} cannot be processed correctly.");
```
JaXt0r, line 143: Why? Do we not support mms? Please add the "why" into the log message.

JaXt0r (changes requested):

Assets/GothicVR/Scripts/Creator/VobCreator.cs
JucanAndreiDaniel: Done.

Assets/GothicVR/Scripts/Creator/MeshCreator.cs
JucanAndreiDaniel: Moved the return to before the GO is created so we don't have to clean it up :D

JaXt0r (approved):
LGTM. Thanks.

PR #142, opened 2023-07-10
Fixing teleport system by adding it after world mesh is loaded into scene

JaXt0r:
# Checklist

## Scene

# Testing


JaXt0r:
Please test on mobile VR if it's working too.

JucanAndreiDaniel (approved):
LGTM

PR #147, opened 2023-07-15
Feature/AudioManager

JucanAndreiDaniel:
Implement Audio Manager which handles sounds, enable/disable the gameobject of the sound based on which grid it's positioned in. To check if a GameObject should be audible to the player we assign them into cells in a grid, and based on where the player is in the grid we enable of disable the gameobject.

# Checklist

## Scene
* [x] FeatureFlags ~~reverted~~ changed to right value

# Testing


Assets/GothicVR/Scripts/Creator/SoundCreator.cs
```diff
@@ -150,5 +162,20 @@ private AudioSource CreateAndAddAudioSource(GameObject soundObject, string sound
...
+        {
+            SetPosAndRot(obj, position.ToUnityVector(), rotation.ToUnityMatrix().rotation);
+        }
+
+        private void SetPosAndRot(GameObject obj, UnityEngine.Vector3 position, Quaternion rotation)
+        {
+            // FIXME - This isn't working
+            if (position.Equals(default) && rotation.Equals(default))
+                return;
+
```
JaXt0r, line 176:
```suggestion

```

Assets/GothicVR/Scripts/Manager/AudioManager.cs
```diff
@@ -0,0 +1,127 @@
...
+            if (listenerCell == playerCell)
+            {
+                return;
+            }
+
+            playerCell = listenerCell;
+
+            // Iterate over the cells around the player in a circular area
+            int radius = 3;
+            for (int x = listenerCell.x - radius; x <= listenerCell.x + radius; x++)
```
JaXt0r, line 51: Please put this calculation in a separate function and add an extensive comment. Seems quite complex from a first read. Feel free to add some links where you have the implementation idea from.

Assets/GothicVR/Scripts/Manager/AudioManager.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            gameObj.SetActive(isAudible);
+        }
+
+        public void SetAudioListener(AudioListener audioListener)
+        {
+            this.audioListener = audioListener;
+        }
+
+        private Vector3Int GetGridCellFromPosition(Vector3 position)
```
JaXt0r, line 109: Please add references (which hopefully include diagrams of this 3d grid idea) for easy understanding.

Assets/GothicVR/Scripts/Manager/AudioManager.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            int x = Mathf.FloorToInt(position.x / cellSize);
+            int y = Mathf.FloorToInt(position.y / cellSize);
+            int z = Mathf.FloorToInt(position.z / cellSize);
+            return new Vector3Int(x, y, z);
+        }
+
+        private bool ShouldBeAudible(GameObject gameObj)
+        {
+            if (audioListener == null)
```
JaXt0r, line 119: Needed?

JaXt0r (changes requested):
Oh my... Another big big PR. :-D
Happy reviewing.

Assets/GothicVR/Scripts/Creator/VobCreator.cs
```diff
@@ -203,8 +202,7 @@ private void CreateSoundDaytime(PxVobSoundDaytimeData vob)
             if (!FeatureFlags.I.EnableSounds)
                 return;
             
-            var vobObj = soundCreator.Create(vob, parentGos[vob.type]);
-            SetPosAndRot(vobObj, vob.position, vob.rotation!.Value);
```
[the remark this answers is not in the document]
JucanAndreiDaniel: We need the position to be set before adding the item to the grid, so that we can actually see in which grid cell the gameobject is
if we do it after the gameobject is created, the gameobject will be set at cell 0,0,0

Assets/GothicVR/Scripts/Manager/AudioManager.cs
JucanAndreiDaniel: Decoupled the calculation to separate function UpdateAudioSourcesWithinRadius with added summary

Assets/GothicVR/Scripts/Manager/AudioManager.cs
JucanAndreiDaniel: added summary for function, vizualisation is in the summary of UpdateAudioSourcesWithinRadius

Assets/GothicVR/Scripts/Manager/AudioManager.cs
JucanAndreiDaniel: nope, removed

Assets/GothicVR/Scripts/Manager/AudioManager.cs
```diff
@@ -0,0 +1,127 @@
...
+        private Dictionary<Vector3Int, List<GameObject>> gridCells = new Dictionary<Vector3Int, List<GameObject>>();
+        private float cellSize = 50f; // Cell size in meters
+
+        private void Create()
+        {
+            // Initialize the grid cells
+            gridCells.Clear();
+        }
+
+        private void Update()
```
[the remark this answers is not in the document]
JucanAndreiDaniel: performance seems to be the same for both approaches, but opted to have the function called once every half a second, in case it would have a performance hit on quest/pico for multiple sounds sources if it were executed every frame

Assets/GothicVR/Scripts/Manager/AudioManager.cs
```diff
@@ -0,0 +1,148 @@
...
+            int y = Mathf.FloorToInt(position.y / cellSize);
+            int z = Mathf.FloorToInt(position.z / cellSize);
+            return new Vector3Int(x, y, z);
+        }
+
+        /// <summary>
+        /// Updates the audio sources within a specified radius around the listener cell.
+        /// It iterates over each cell within the radius and checks if it falls within the circular area.
+        /// If so, it retrieves the audio sources associated with that cell and determines if they should be audible.
+        /// https://demonstrations.wolfram.com/ApproximatingSpheresWithBoxes/ - graphical visualisation of how the sphere would look like
```
JaXt0r, line 107: Thank you. This helps.

JaXt0r, line 107: Would it make sense to have size of the boxes configurable? i.e. how do you get the most performance out of it?

JaXt0r (approved):
Thank you for the hint to WolframAlpha.

Assets/GothicVR/Scripts/Manager/AudioManager.cs
JucanAndreiDaniel: IMO most performance would be to not have many cells to check, in this case it would be 3^3 = 27

PR #150, opened 2023-07-17
Feature/world async

JucanAndreiDaniel:
Async Loading, no more game is not responding until you spawn, now it's a ~~black screen (at the time of writing)~~ loading screen until you spawn, also works when traveling between worlds

# Checklist

## Scene

# Testing
* [ ] Tested with PCVR and one mobile VR device (Pico/Quest)


Assets/GothicVR/Editor/OcclusionCullingTool.cs
```diff
@@ -55,7 +55,7 @@ public static void ShowWindow()
             GameData.I.VdfsPtr = _vdfsPtr;
             
             // use PhoenixImporter to handle loading the world and setting it to the correct scene.
-            WorldCreator.I.LoadEditorWorld(_vdfsPtr, "world");
+            // WorldCreator.I.LoadEditorWorld(_vdfsPtr, "world");
```
JaXt0r, line 58: Why commenting it out? Please redo it comment at least why.

Assets/GothicVR/Scenes/Loading.unity
JaXt0r, line 1: Please update .drawio diagram as the loading scene will be before world loaded I assume.

Assets/GothicVR/Scripts/Caches/AssetCache.cs
```diff
@@ -62,6 +63,37 @@ public Texture2D TryGetTexture(string key)
             return texture;
         }
 
+        public async Task<Texture2D> TryGetTextureAsync(string key)
```
JaXt0r, line 66: Uh really? Why do we need to start with async stuff? If needed, we should rethink where to put the async calls.

Assets/GothicVR/Scripts/Creator/MeshCreator.cs
```diff
@@ -52,23 +58,38 @@ public GameObject Create(WorldData world, GameObject parent)
...
 
-                PrepareMeshRenderer(meshRenderer, subMesh);
+                await PrepareMeshRenderer(meshRenderer, subMesh);
+
                 PrepareMeshFilter(meshFilter, subMesh);
                 PrepareMeshCollider(subMeshObj, meshFilter.mesh, subMesh.material);
 
                 subMeshObj.SetParent(meshObj);
+
+                // Update the sub-mesh creation progress
```
JaXt0r, line 68: Please create a LoadingManager or ProgressbarManager so that we can have a Singleton which is handling the math within some functions.

This will lower our code on this class

Assets/GothicVR/Scenes/Loading.unity
JaXt0r, line 1: Two scenes have the VRPlayer attached to it. Please fix.
image

Assets/GothicVR/Scripts/Manager/UIManager.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    public class UIManager : SingletonBehaviour<UIManager>
+    {
+        public Material GothicLoadingMenuMaterial;
+        public Material LoadingBarBackgroundMaterial;
+        public Material LoadingBarMaterial;
+        public void LoadDefaultTextures()
+        {
+            var loadingBackgroundTexture = AssetCache.I.TryGetTexture("LOADING.TGA");
+            GothicLoadingMenuMaterial.mainTexture = loadingBackgroundTexture;
```
JaXt0r, line 21: Now I get it with the materials being used.
Nevertheless I would challenge one more thing: Do the materials differ on their settings in the Asset folder? If so, how much? i.e. if there are more than 3 settings you need to do for the materials, we should definitely have them on assets. If not... I will probe again. ;-)

Assets/GothicVR/Materials/SphereLoading.mat
JaXt0r, line 1: Rename to have Prefix called Loading* as the other materials are.

JaXt0r, line 1: Please also put the Loading materials inside a /Loading/ folder. Then you could also get rid of the duplicate prefixes in Loading*.mat

Assets/GothicVR/Scenes/Bootstrap.unity
```diff
@@ -865,3 +866,19 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   dontDestroyOnLoad: 0
   interactionManager: {fileID: 1300502570}
+--- !u!114 &1896514216
+MonoBehaviour:
```
JaXt0r, line 870: When do we actually add the Loading.unity scene? Is it done after Phoenix is bootstrapped? If so, we could think of having this Manager and its materials "not" attached to Bootstrap, but to its own Loading.unity scene. And load the textures once the Loading.unity scene is loaded.

Assets/GothicVR/Scripts/Bootstrap/PhoenixBootstrapper.cs
```diff
@@ -115,6 +115,11 @@ private void LoadMusic()
             Debug.Log("Loading music");
         }
 
+        private void LoadTextures()
+        {
+            UIManager.I.LoadDefaultTextures();
```
JaXt0r, line 120: As I said before: If we load Loading.unity scene after bootstrapping is done (Which I prefer), then we can remove this one here and have it inside Start() of the Manager on the Loading scene.

Assets/GothicVR/Scripts/Creator/VobCreator.cs
```diff
@@ -40,34 +41,39 @@ private void Start()
...
-                
+
                 parentGos.Add(type, newGo);
             }
         }
 
-        private void CreateVobs(GameObject root, PxVobData[] vobs)
+        private async Task CreateVobs(GameObject root, PxVobData[] vobs, Action<float> progressCallback, float progress = 0.5f)
         {
+            float totalProgress = progress;
```
JaXt0r, line 70: Why not putting it into a LoadingManager? I think of the following functions:
1. AddSegment(steps) --> LoadingSegment
2. LoadingSegment.addIncrement() // which will automatically update the loading bar UI.
(First idea. Please challenge it)

Cool thing about it: You can easily reuse it on MeshCreator etc.

Assets/GothicVR/Scripts/Creator/WorldCreator.cs
```diff
@@ -30,13 +32,15 @@ public void Create(string worldName)
 
             GameData.I.WorldScene!.Value.GetRootGameObjects().Append(worldGo);
 
-            worldMesh = MeshCreator.I.Create(world, worldGo);
-            VobCreator.I.Create(worldGo, world);
+            worldMesh = await MeshCreator.I.Create(world, worldGo, progress => progressCallback?.Invoke(progress * 0.5f));
```
JaXt0r, line 35: Ah we really need to see how this works. From JavaScript I know, that await is just making an async function synchronous. So no additional benefits. Let's check on it at the end.

Assets/GothicVR/Scripts/Creator/WorldCreator.cs
```diff
@@ -185,7 +189,7 @@ public void LoadEditorWorld(IntPtr vdfPtr, string zen)
             sampleScene.GetRootGameObjects().Append(worldGo);
 
             // load only the world mesh
-            SingletonBehaviour<MeshCreator>.GetOrCreate().Create(world, worldGo);
+            await SingletonBehaviour<MeshCreator>.GetOrCreate().Create(world, worldGo, progress => { });
```
JaXt0r, line 192:
```suggestion
            await MeshCreator.I.Create(world, worldGo, progress => { });
```

Why are you putting a progress function into?

Assets/GothicVR/Scripts/Demo/LoadFollower.cs
```diff
@@ -0,0 +1,36 @@
...
+    private Vector3 initialOffset; // Initial offset between the logo and the camera
+    private Vector3 velocity = Vector3.zero; // Velocity for smoothing the movement
+
+    private void Start()
+    {
+        // Calculate and store the initial offset between the logo and the camera
+        initialOffset = transform.position - target.position;
+    }
+
+    private void LateUpdate()
```
JaXt0r, line 18: Why is late Update needed? Please comment if so.

Assets/GothicVR/Scripts/Demo/LoadFollower.cs
```diff
@@ -0,0 +1,36 @@
...
+    public Transform target; // Reference to the main camera
+    public float smoothness = 0.1f; // Smoothing factor for the logo movement
+    public float followDelay = 0.5f; // Delay between the logo and the main camera
+
+    private Vector3 initialOffset; // Initial offset between the logo and the camera
+    private Vector3 velocity = Vector3.zero; // Velocity for smoothing the movement
+
+    private void Start()
+    {
+        // Calculate and store the initial offset between the logo and the camera
```
JaXt0r, line 14: Thanks for the comments. This helps understanding.

Assets/GothicVR/Scripts/Manager/GvrSceneManager.cs
```diff
@@ -26,80 +32,146 @@ public class GvrSceneManager : SingletonBehaviour<GvrSceneManager>
         /// </summary>
         public void LoadStartupScenes()
         {
-            generalScene = SceneManager.LoadScene(generalSceneName, new LoadSceneParameters(LoadSceneMode.Additive));
-            SceneManager.MoveGameObjectToScene(interactionManager, generalScene);
-
-            LoadWorld("world.zen", "FP_PICKRICE_SWAMP_02");
+            LoadWorld("oldmine.zen", "ENTRANCE_SURFACE_OLDMINE");
```
JaXt0r, line 35: Can you please add a FeatureFlag-property where we have a dropdown of ~5 well known spawn points? This would be better than hard coding it here.

Assets/GothicVR/Scripts/Manager/GvrSceneManager.cs
```diff
@@ -26,80 +32,146 @@ public class GvrSceneManager : SingletonBehaviour<GvrSceneManager>
...
-            var newWorldScene = SceneManager.LoadScene(worldName, new LoadSceneParameters(LoadSceneMode.Additive));
+            var asyncLoad = SceneManager.LoadSceneAsync(worldName, new LoadSceneParameters(LoadSceneMode.Additive));
+            Scene newWorldScene;
 
             // Remove previous scene.
             // TODO - it might be, that we need to wait for old map to be removed before loading new one. Let's see...
             if (GameData.I.WorldScene.HasValue)
             {
+                // Try to delete everything in old scene, so we can load everything again from scratch
+                var objects = GameData.I.WorldScene.Value.GetRootGameObjects();
```
JaXt0r, line 52: We don't need to destroy objects from scene manually, do we? What happens if we just remove the scene? Everything should be gone. I don't want to iterate through the objects, please.

Assets/GothicVR/Scripts/Manager/GvrSceneManager.cs
```diff
@@ -26,80 +32,146 @@ public class GvrSceneManager : SingletonBehaviour<GvrSceneManager>
...
+                {
+                    GameObject.Destroy(objects[i]);
+                }
                 SceneManager.UnloadSceneAsync(GameData.I.WorldScene.Value);
             }
 
-            GameData.I.WorldScene = newWorldScene;
+            asyncLoad.completed += async (asyncOperation) =>
+            {
+                newWorldScene = SceneManager.GetSceneByName(newWorldName);
```
JaXt0r, line 62: I think you get the scene from the completed function as parameter. Please check it. You don't need to do a search again afaik.

Assets/GothicVR/Scripts/Manager/GvrSceneManager.cs
```diff
@@ -26,80 +32,146 @@ public class GvrSceneManager : SingletonBehaviour<GvrSceneManager>
...
+            // TODO: for new game we need to load texture "LOADING.TGA"
+                var textureString = "LOADING_" + worldName.Split('.')[0].ToUpper() + ".TGA";
+                UIManager.I.setTexture(textureString, UIManager.I.GothicLoadingMenuMaterial);
+            }
+
+            var loadingScene = SceneManager.LoadScene("Loading", new LoadSceneParameters(LoadSceneMode.Additive));
+
+            SceneManager.sceneLoaded += (Scene scene, LoadSceneMode mode) =>
+            {
+                if (scene == loadingScene)
```
JaXt0r, line 132: Ah I see.

To be honest I don't really like these inline events. They can make it hard to read. Please check if it would make sense to have separate methods subscribed. Then we can easily spot what is called and when.

JaXt0r (changes requested):
What a big PR... :-D

Assets/GothicVR/Scenes/Loading.unity
JucanAndreiDaniel: This was by intention, the General scene contains the general VR player, the Loading scene contains a "nerfed" down version without most of the stuff and the canvas for loading stuff attached to it (structure came from previous iteration when the canvas followed the player camera as to have the loading menu in front of the player all the time)

Assets/GothicVR/Scripts/Manager/UIManager.cs
JucanAndreiDaniel: Makes sense, we can create the material in script, seems to be a better idea

Assets/GothicVR/Scripts/Creator/WorldCreator.cs
JucanAndreiDaniel: using await on a function makes the parent function "wait" for our function to finish, but it runs in a non-blocking way for Unity as it should handle everything on the main thread, so instead of blocking the main thread and render the world, we can do async stuff (do small snippets) without blocking the main thread, thus making the loading screen feasible

Assets/GothicVR/Scripts/Creator/WorldCreator.cs
JucanAndreiDaniel: empty action callback that will be removed when I do the LoadingManager stuff

Assets/GothicVR/Scripts/Demo/LoadFollower.cs
JucanAndreiDaniel: This was so we can add a delay to when the gameobject this script was attached to (loading canvas) was moved based on the transform specified.
will add a summary to it if this will still be used

Assets/GothicVR/Scripts/Manager/GvrSceneManager.cs
JucanAndreiDaniel: This was more of a test that slipped up since as i mentioned on Discord, the gameobjects per scene are cached after being loaded once as so when we go to a world for a second time, instead of loading again the object from gothic files, we get them from memory in a blocking way, which i wanted to avoid for UX reasons (we want a loading screen between worlds, not to be stuck and load into a world without feedback on what's going on)

Assets/GothicVR/Scripts/Creator/MeshCreator.cs
JucanAndreiDaniel: Introduced LoadingManager to handle storing progress and updating the UI to reflect the changes

Assets/GothicVR/Scripts/Manager/UIManager.cs
JucanAndreiDaniel: Removed the materials and using script generated materials

Assets/GothicVR/Scripts/Manager/GvrSceneManager.cs
JucanAndreiDaniel: removed the async loading

Assets/GothicVR/Scenes/Loading.unity
JucanAndreiDaniel: restructured to have the canvas be a child of the loading sphere (this is just so we have a background), i'd still like to have the VR player as it is.

JaXt0r: Wait. Isn't this more complex if you have a prefab twice in two scenes with different settings inside?

Please think of it:
* What do you gain by having the Player twice?
* Is it worth the duplicate maintenance?
* What happens if we load another scene? Will we need to remove the other player in General Scene and load it again?

JucanAndreiDaniel: Well it's not really more complex as i treat the prefab like a template and modifying something that is removed from local instance shouldn't affect the local copy.
1. we don't have to mess much with the position of the player from General scene to disable gravity and movement
2. mentioned in the beginning
3. when loading another world scene we unload the General scene along with the old world scene, loading the Loading scene and then preparing the new world scene, when everything is finished then we remove the loading scene and load again General scene and position the player properly

Assets/GothicVR/Scripts/Manager/LoadingManager.cs
```diff
@@ -0,0 +1,96 @@
...
+        public void AddProgress(LoadingProgressType progressType, float progress)
+        {
+            float newProgress = progressByType[progressType] + progress;
+            progressByType[progressType] = newProgress;
+            UpdateLoadingBar();
+        }
+
+        public void SetMaterialForLoading(Scene scene)
+        {
+            scene.GetRootGameObjects().FirstOrDefault(go => go.name == "LoadingSphere").transform.Find("LoadingCanvas/LoadingImage/ProgressBackground/ProgressBar").gameObject.GetComponent<Image>().material = UIManager.I.LoadingBarMaterial;
```
JaXt0r, line 90: I see your point in referencing the scene elements via path. But please think of:
1. Couldn't it be better to have just the last element and do the find? We have DeepFindChildren already in main branch and it could be used. (It's called GameObject.FindChildRecursive()). Because if we change the path sometime, it will all fall apart.
2. If you really want to stick that way, then please name the paths on top as private const variables. (e.g. you already name ProgressBar path twice, which would benefit from your contanciation :-D )

JaXt0r (changes requested):
Hi Frank. Did a review on your current version. I have three major concerns where we should sync about.

1. The code for MeshCreator and others changed a lot in main branch. I'm afraid you'll have to do a big merge conflict fix.
2. But before you fetch main and start it: Let's please check again if it's sufficient to only have the "entry" method marked as _async_ I'm not 100% sure, but for me it seems we need to have 1 method marked as async and call it normal. Then it runs in a parallel Thread. No other sub-methods need to be marked async. Please advice.
3. I'm still not on board about having Loading scene being there separately and duplicate VRPlayer. Please think about if it makes sense to add Loading.unity into General.unity and just disable it when world loads. If not, please explain why.

JucanAndreiDaniel:
> Hi Frank. Did a review on your current version. I have three major concerns where we should sync about.
> 
>     1. The code for MeshCreator and others changed a lot in main branch. I'm afraid you'll have to do a big merge conflict fix.
> 
>     2. But before you fetch main and start it: Let's please check again if it's sufficient to only have the "entry" method marked as _async_ I'm not 100% sure, but for me it seems we need to have 1 method marked as async and call it normal. Then it runs in a parallel Thread. No other sub-methods need to be marked async. Please advice.
> 
>     3. I'm still not on board about having Loading scene being there separately and duplicate VRPlayer. Please think about if it makes sense to add Loading.unity into General.unity and just disable it when world loads. If not, please explain why.

on the second point: i left only mesh creation to be done async for now, and it needs to be deeply rooted into the GetTexture function as Unity will absolutely not create a texture if its not on the main thread

on the 3rd point: i moved everything from Loading Scene to General scene, so we have 2 root game objects: PlayerController and LoadingSphere. We leave active only the root gameobject which makes sense, loading sphere with the very cut down vr player or the player controller which is the normal player.

Why i want to use the cut down vr player for loading stuff is because i dont want to mess in script to disable locomotion and settings menu stuff and re-enable them after world is loaded. For the scene stuff i have no problems merging them but i dont want to overcomplicate the player stuff by messing with the components in scripts when we can very easily have 2 vr players which serve different purposes (normal gameplay and spectator, no nothing except looking around)

JucanAndreiDaniel:
> on the second point: i left only mesh creation to be done async for now, and it needs to be deeply rooted into the GetTexture function as Unity will absolutely not create a texture if its not on the main thread

For this i mean that doing something like this `Task.Run(() => {var texture = new Texture2()});`

JaXt0r (changes requested):
Tested.
* Loading works fine when starting the game. Nicely done!
* When moving back from oldmine to world, I see the loading screen briefly, but SteamVR says "Wait" and opens up some SteamVR overlay. This is normally the case when a game is unresponsive. Question: Is the async still working when moving between scenes?
* Performance is super bad. Especially when moving to Oldcamp. Is this just my PC and Editor build or do you experience it as well?

Most importantly:
* We need to be super careful with async-await.
* I expect (but my current tests failed) to have only 1 single method with async and 1 yield await. Nothing more. The sub-methods are automatically running in second thread. No further async-await needed. I put a diagram of my understanding at the end.
* I checked documentation further on how to create an async function without real await. We just need to call await Task.Yield(); - which is like coroutine usage neccessarity.

Please check and challenge my thoughts on async usage with your knowledge.

image

JaXt0r:
I'm closing it as we reworked the solution here: https://github.com/GothicVRProject/GothicVR/pull/163

PR #151, opened 2023-07-19
Feature/10.2 npc animations

JaXt0r:
# Checklist

## Scene

# Testing

## Dependencies


JucanAndreiDaniel (approved):
LGTM, tested on a quest 2 and it didn't crash when loading directly in old camp with npc's

PR #153, opened 2023-07-20
Optimized SingletonBehaviour and removed unneccessary methods.

JaXt0r:
Just changed one of our most used class as it had a lot of unneccessary stuff in there.


JucanAndreiDaniel (approved):
LGTM

PR #158, opened 2023-07-23
Feature/kiss the cook

JaXt0r:
# To test
1. Start game
2. Check if NPCs have different heads and body color
3. Check if Snaf is there (Hint: His head is slightly misplaced, but I will fix this later)

# Checklist

## Scene

# Testing
* [x] Tested with PCVR


Assets/GothicVR/Scripts/Creator/Meshes/NpcMeshCreator.cs
```diff
@@ -0,0 +1,68 @@
...
+            if (name.ToUpper().Contains("MOUTH") || name.ToUpper().Contains("TEETH"))
+                return base.GetTexture(name);
+            
+            if (!name.ToUpper().EndsWith("V0_C0.TGA"))
+            {
+                Debug.LogError($"The format of body texture isn't right for ${name}");
+                return base.GetTexture(name);
+            }
+
+            var formattedTextureName = Regex.Replace(name, "(?<=.*?)V0_C0", $"V{tmpBodyTexNr}_C{tmpBodyTexColor}");
```
JucanAndreiDaniel, line 63: Can you add a comment on what the regex does? the first part might be confusing for people not very knowledgeable about regex

JucanAndreiDaniel (approved):
LGTM, just regex part needs explanation

Assets/GothicVR/Scripts/Creator/Meshes/NpcMeshCreator.cs
JaXt0r: Done. Thanks for the hint.

Romeeey:
Did you document (eg in a ticket) what the exact mechanism with the v and c suffix exactly is? 
I would recommend to do so.

Implementation looks good.

PR #159, opened 2023-07-24
Feature/playerprefs menu

Romeeey:
# Checklist
* [ ] Open settings menu (right controller, primary button)
* [ ] Change from snapturn to continuous turn
* [ ] Increase move speed on slider
* [ ] Close the app (PCVR)
* [ ] Restart the app 
* [ ] Check that continuous turn and the high speed are still in place
Tested on Quest 2 with PCVR

# Testing
* [x] Tested with PCVR


JaXt0r:
Tested. I experienced some bugs and improvement ideas:
1. bug: I spawm inside water and can't go up. (We had this before, but I don't remember what it was...)
2. idea: The menu floats on top of me. How about we have it fixed in the world and once we open it, it won't move, but we can move around it?
3. idea: The menu opens with >A< or >main< button. Please put it to a button which is good for main menu. On Index it could be Button 2 and on Quest Button 1. ([button reference](https://github.com/GothicVRProject/GothicVR_PoC/blob/main/Documentation/How-Tos/How-To-Play.md)) - Whatever it is we should align somehow with other games and which button they use to open up a menu.

JaXt0r (changes requested):
Some improvements needed.

Romeeey:
> Tested. I experienced some bugs and improvement ideas:
> 
> 1. bug: I spawm inside water and can't go up. (We had this before, but I don't remember what it was...)

That seems to be a general bug / problem that happens from time to time. The main camera loses some parameters, when turning on play mode. If you copy over these values from the main camera of basic xr origin (tyroonis) it's ok. It's not branch / commit specific, this happens local when entering playmode.

image

I don't really know why this is happening.

> 2. idea: The menu floats on top of me. How about we have it fixed in the world and once we open it, it won't move, but we can move around it?

It is implemented that way. The menu is always in front of you, when hidden. When it is activated it gets visible and stays fixed in the world. You can now come further or move away or walk around.

> 3. idea: The menu opens with >A< or >main< button. Please put it to a button which is good for main menu. On Index it could be Button 2 and on Quest Button 1. ([button reference](https://github.com/GothicVRProject/GothicVR_PoC/blob/main/Documentation/How-Tos/How-To-Play.md)) - Whatever it is we should align somehow with other games and which button they use to open up a menu.

I chose a random button for now, since we haven't decided yet on a set of input buttons. In the PoC we didn't have a dedicated button, but a main menu that had a settings GUI button.



> 
>

Assets/GothicVR/Scripts/UI/TurnSettingDropdownController.cs
```diff
@@ -21,31 +21,41 @@ void Awake()
...
+            case 0:
             default:
-                snapTurn.enabled = true;
-                continuousTurn.enabled = false;
+                EnableSnapTurn();
                 break;
         }
     }
+
+    void EnableSnapTurn()
```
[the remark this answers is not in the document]
Romeeey: Cool.

Assets/GothicVR/Scripts/UI/MoveSpeedController.cs
```diff
@@ -12,10 +12,12 @@ void Start()
     {
         Slider speedslider = transform.GetComponent<Slider>();
         speedslider.onValueChanged.AddListener(ChangeMoveSpeed);
+        speedslider.value = PlayerPrefs.GetFloat("MoveSpeed", 8f);
```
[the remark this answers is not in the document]
Romeeey: Can do.

Assets/GothicVR/Scripts/UI/TurnSettingDropdownController.cs
```diff
@@ -21,31 +21,41 @@ void Awake()
...
             dropdown.options.Add(new TMP_Dropdown.OptionData() { text = item });
-
         }
         dropdown.itemText.font = GameData.I.GothicSubtitleFont;
-
         dropdown.onValueChanged.AddListener(delegate { DropdownItemSelected(dropdown); });
         snapTurn = locomotionsystem.GetComponent<ActionBasedSnapTurnProvider>();
         continuousTurn = locomotionsystem.GetComponent<ActionBasedContinuousTurnProvider>();
+
+        dropdown.value = PlayerPrefs.GetInt("TurnSetting");
```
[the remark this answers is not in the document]
Romeeey: Done

Assets/GothicVR/Scripts/UI/TurnSettingDropdownController.cs
JaXt0r, line 1: You can simplify the dropdownChangeEvent. I added it as suggestions for you.

Assets/GothicVR/Scripts/UI/TurnSettingDropdownController.cs
```diff
@@ -21,31 +22,41 @@ void Awake()
...
+        snapTurn.enabled = true;
+        continuousTurn.enabled = false;
+        PlayerPrefs.SetInt("TurnSetting", 0);
+    }
+
+    void EnableContinuousTurn()
+    {
+        snapTurn.enabled = false;
+        continuousTurn.enabled = true;
+        PlayerPrefs.SetInt("TurnSetting", 1);
```
JaXt0r, line 60: Constant please.

Assets/GothicVR/Scripts/UI/TurnSettingDropdownController.cs
```diff
@@ -21,31 +22,41 @@ void Awake()
...
+                EnableSnapTurn();
                 break;
         }
     }
+
+    void EnableSnapTurn()
+    {
+        snapTurn.enabled = true;
+        continuousTurn.enabled = false;
+        PlayerPrefs.SetInt("TurnSetting", 0);
```
JaXt0r, line 53: Constant please.

We can also think about a whole file where we only store the Constants.

e.g. 
```c#
public class PlayerPrefConstants
```
Because it will be in the future, that we need to get these keys somewhere else if we stick with PlayerPref.

JaXt0r (changes requested):
A few things to notice:
1. Performance is so super bad. I can barely even open my SteamVR overlay. Just want to mention it as it's unplayable unfortunately. Also on your end?
2. Klicking >A< again won't close this window. Please implement as I moved away from this screen and didn't see it is still open.
3. The menu is only visible from front. But if you walk around it, the text is still visible, but not the main texture.

I recommend at least implementing topic 2. ;-)

Assets/GothicVR/Scripts/UI/TurnSettingDropdownController.cs
Romeeey: Done.

Assets/GothicVR/Scripts/UI/TurnSettingDropdownController.cs
Romeeey: Added to constants manager. done.

Romeeey:
> A few things to notice:
> 
> 1. Performance is so super bad. I can barely even open my SteamVR overlay. Just want to mention it as it's unplayable unfortunately. Also on your end?
> 2. Klicking >A< again won't close this window. Please implement as I moved away from this screen and didn't see it is still open.
> 3. The menu is only visible from front. But if you walk around it, the text is still visible, but not the main texture.
> 
> I recommend at least implementing topic 2. ;-)

As already discussed let's implement 2 and 3. But I will create a new branch since we need to change our input controller to fully support action based system. This is a bigger change.

JaXt0r (approved):
LGTM

PR #161, opened 2023-07-25
add position and rotation again to vr player

JucanAndreiDaniel:
Fix the vr player not tracking again the hmd position and rotation


JaXt0r (approved):
Works for my Valve Index.

PR #166, opened 2023-07-30
Update test_and_build.yml

JaXt0r:
Build only on labelling from now on.

# To test
1. Check if there is a pipeline run named below (should be 0)
2. Add label "Pipeline"
3. Check if pipeline runn will start
4. Once done, please see if there are binaries created in "Actions" tab


JaXt0r:
Ah! We need to merge it to main before it works actually. 🤦

PR #171, opened 2023-07-31
Bugfix/climbable ladders

### LINKED ISSUE #170: Ladder climbing not working
Tested on main branch, cannot climb ladders

JucanAndreiDaniel:
# To test
1. ladder if it is climbable

# Testing
* [x] Tested with PCVR

closes #170

### DIFF
--- a/Assets/GothicVR/Resources/Prefabs/UI Element Prefabs/VRPlayer.prefab
+++ b/Assets/GothicVR/Resources/Prefabs/UI Element Prefabs/VRPlayer.prefab
@@ -626,8 +626,8 @@ GameObject:
   m_Component:
   - component: {fileID: 3286136461689514590}
   - component: {fileID: 3286136461689514587}
-  - component: {fileID: 3286136461689514588}
   - component: {fileID: 3286136461689514589}
+  - component: {fileID: 7127237056583641875}
   m_Layer: 7
   m_Name: LeftHandBaseController
   m_TagString: Untagged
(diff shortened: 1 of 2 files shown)


JaXt0r (commented):
Can you please explain:
1. What are the changes you did on the Prefab?
2. Was there a mesh collider before? Wouldn't it be enough the collider is on ladders only? I have a feeling we put Colliders on too much Elements now.

JucanAndreiDaniel:
> Can you please explain:
> 
>     1. What are the changes you did on the Prefab?
> 
>     2. Was there a mesh collider before? Wouldn't it be enough the collider is on ladders only? I have a feeling we put Colliders on too much Elements now.

1. We lost the custom XR Direct Interactor for climbing and i modifed the prefab to include it for hands
https://github.com/GothicVRProject/GothicVR/blob/main/Assets/GothicVR/Scripts/Player/Climb/XRDirectClimbInteractor.cs
2. There was mesh collider for ladders, doors and any object created from attachments. If there is no mesh for ladders then we cannot interact with it
I'm not sure if we want people to walk through doors, torches (e.g. the torch outside the oldcamp entrance towards oldmine) or even the big apparatus in front of the entrance of oldmine

JaXt0r (approved):
Now I get it. For boned meshes, we sometimes have attachments (no bones, but fixed meshes on an object.)

And here it makes definitely sense to have colliders. Good fix!

PR #175, opened 2023-08-03
Feature/ladder teleport

JucanAndreiDaniel:
Implementing a first version of teleport up and down a ladder (since we don't want to climb them manually anymore)

# To test
1. Go to a ladder
1.1. Hold 0.5 seconds the teleport ray on the ladder
1.2. Hold and grab for 0.5 seconds with the right ray hand (ray hands for menu)

# Testing
* [x] Tested with PCVR

### DIFF
--- a/Assets/GothicVR/Scripts/Player/Climb/RayTeleportClimb.cs
+++ b/Assets/GothicVR/Scripts/Player/Climb/RayTeleportClimb.cs
@@ -83,7 +83,8 @@ public class RayTeleportClimb : MonoBehaviour
         isHittingObject = false;
 
         // Deactivate the teleport ray
-        interactor.enabled = false;
+        if (interactor.gameObject.name.Contains("Teleport"))
+            interactor.enabled = false;
     }
 
     private void TeleportPlayer(Vector3 targetPosition)
--- a/Assets/GothicVR/Scripts/Player/Climb/RayTeleportClimb.cs
+++ b/Assets/GothicVR/Scripts/Player/Climb/RayTeleportClimb.cs
@@ -10,7 +10,6 @@ public class RayTeleportClimb : MonoBehaviour
     [SerializeField] private GameObject player;
 
     private bool isHittingObject = false;
-    private GameObject hitObject;
     private Vector3 zsPos0Position;
     private Vector3 zsPos1Position;
     private float hitTime;
@@ -31,7 +30,7 @@ public class RayTeleportClimb : MonoBehaviour
         if (args.interactableObject != null && args.interactableObject.transform.CompareTag(ConstantsManager.I.ClimbableTag))
         {
             // Show a message in the logs
-            hitObject = args.interactableObject.transform.gameObject;
+            var hitObject = args.interactableObject.transform.gameObject;
 
             // Get the zs_pos0 and zs_pos1 positions
             zsPos0Position = hitObject.FindChildRecursively(zsPos0).transform.position;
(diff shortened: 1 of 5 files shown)


Assets/GothicVR/Scripts/Player/Climb/RayTeleportClimb.cs
```diff
@@ -0,0 +1,90 @@
...
+    private void Start()
+    {
+        interactor.selectEntered.AddListener(OnRaycastHit);
+        interactor.selectExited.AddListener(OnRaycastExit);
+    }
+
+    private void OnRaycastHit(SelectEnterEventArgs args)
+    {
+        // Check if the interactable GameObject has the tag "Climbable"
+        if (args.interactableObject != null && args.interactableObject.transform.CompareTag("Climbable"))
```
JaXt0r, line 27: Please use our constant for it.

Assets/GothicVR/Scripts/Player/Climb/RayTeleportClimb.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            // Show a message in the logs
+            hitObject = args.interactableObject.transform.gameObject;
+
+            // Get the zs_pos0 and zs_pos1 positions
+            zsPos0Position = hitObject.FindChildRecursively("ZS_POS0").transform.position;
```
JaXt0r, line 33: This is so cool.
Please Put these as constants on top.

Assets/GothicVR/Scripts/Player/Climb/RayTeleportClimb.cs
```diff
@@ -0,0 +1,90 @@
...
+    private void PerformTeleport()
+    {
+        // Get the player's position
+        Vector3 playerPosition = player.transform.position;
+
+        float yDifferenceToZsPos0 = Mathf.Abs(playerPosition.y - zsPos0Position.y);
+        float yDifferenceToZsPos1 = Mathf.Abs(playerPosition.y - zsPos1Position.y);
+
+        // Teleport the player to the closer zs_pos position based on y-level
+        if (yDifferenceToZsPos0 < yDifferenceToZsPos1)
```
JaXt0r, line 69: I thought about "if we hit the last x%, then teleport with a blink". How is Alyx doing it?

JucanAndreiDaniel: What do you mean by this? If we climb the ladder a certain percent and if we let go, we should teleport the player to the top/bottom (depending on where the player let go)?

JaXt0r: Like Alyx. When there was only (idk 5%?) a few cm left on top, then it spawns to it. Here I see, that it always spawns which is nearer, right?

JucanAndreiDaniel: We spawn to the "opposite" side, which is farther (compared on the Y axis) as it means that we want to go there, climb up if we are closer to ZS_POS0 (which is down) or climb down if we are closer to ZS_POS1

JaXt0r (approved):
I saw the arrow. Yes, it's not perfect, but it's doing what it should. I would recommend you slow down the animation by 50% (i.e. that teleportation is taking double of the time). But as a bonus. Approved for now. Can't wait to get gamers feedback about it.

PR #181, opened 2023-08-10
bugfix/ladder teleport indicator

JucanAndreiDaniel:
# To test
1. Arrow appears regardless of player position (front/back of ladder)


JaXt0r (approved):
Sure, sure. ;-)

PR #183, opened 2023-08-11
add menu skip flag

JucanAndreiDaniel:
Add flag to skip main menu and load directly into the world
To modify the starting world and position modify selectedWorld and selectedWaypoint in ConstantsManager
# To test
1. Check menu skip flag and load world


JaXt0r (approved):
Cool. I thought about the exact same thing.

PR #189, opened 2023-08-16
Bugfix/raycast hands

Romeeey:
# To test
* [x] Try teleporting with grab raycasts - shouldn't work
* [x] Try remote grab - should work
* [x] Try teleport - should work

# Checklist

## Scene

# Testing
* [ ] Tested with PCVR

## Dependencies


JaXt0r (approved):
Tests worked as mentioned. Teleport can't grab any longer.

PR #191, opened 2023-08-17
Feature/fonts

JucanAndreiDaniel:
# To test
1. Menu font should be right
2. Settings font should be right

# Testing
* [x] Tested with PCVR


Assets/GothicVR/Resources/FontAsset/LiberationSans SDF empty.asset
JaXt0r, line 1: Why is this needed? An empty font or so?
Will it work without it?

Assets/GothicVR/Scripts/Player/Menu/MenuDropdownController.cs
```diff
@@ -33,7 +33,7 @@ private void SetWorldDropdown()
             worlds.Add("Orc Temple", "orctempel.zen");
             worlds.Add("Free Mine", "freemine.zen");
 
-            worldDropdown.itemText.font = GameData.I.GothicSubtitleFont;
+            worldDropdown.itemText.font = GameData.I.EmptyFont;
```
JaXt0r, line 36: Let's be careful. If we set the empty font to another element as well, then we copy by reference and would change for all assignments.

Why not loading all the fonts and referencing them by name etc.?
In other words: How are Daedalus files handle fonts for menu? Do they reference them or just pick a hardcoded one.

Assets/GothicVR/Scripts/UI/FontManager.cs
```diff
@@ -2,36 +2,114 @@
...
-                // On Main Menu text is very small if gothic font is used
-                if(textComponent.tag == ConstantsManager.I.MenuFontTag && GameData.I.GothicMenuFont != null)
+                textComponent.font = GameData.I.EmptyFont;
+
+                // texture is rotated 180 degrees
+                textComponent.rectTransform.localRotation = Quaternion.Euler(180, 0, 0);
+            }
+        }
+
+        public TMP_SpriteAsset LoadFont(string fontName)
```
JaXt0r, line 38: Do you have a mechanism in place, that a font isn't generated twice? I.e. how do you handle caching of this function?

Assets/GothicVR/Scripts/UI/TurnSettingDropdownController.cs
```diff
@@ -29,7 +29,7 @@ void Awake()
         {
             dropdown.options.Add(new TMP_Dropdown.OptionData() { text = item });
         }
-        dropdown.itemText.font = GameData.I.GothicSubtitleFont;
+        dropdown.itemText.font = GameData.I.EmptyFont;
```
JaXt0r, line 32: The more I see it, the more I think "Empty font" is there ring name. We need to get it by a specific name like GetMenuFont(). However G1 is handling it. We want it as well. 😚

Assets/GothicVR/Scripts/UI/FontManager.cs
```diff
@@ -2,36 +2,114 @@
...
+
+            return spriteAsset;
+        }
+
+        static Material GetDefaultSpriteMaterial(Texture2D spriteSheet = null)
+        {
+            ShaderUtilities.GetShaderPropertyIDs();
+
+            // Add a new material
+            Shader shader = Shader.Find("TextMeshPro/Sprite");
```
JaXt0r, line 108: Do we have a cache for Shader lookup already? If not, we should do it. We can use the LookupCache.cs file which is for that.

JaXt0r (changes requested):
A few questions only.

Assets/GothicVR/Resources/FontAsset/LiberationSans SDF empty.asset
JucanAndreiDaniel: It is needed, this is a font without any characters, which makes TMP fall back to the spriteAsset which contains the actual font

Assets/GothicVR/Scripts/Player/Menu/MenuDropdownController.cs
JucanAndreiDaniel: For now we only use one font, this remains as is, when we want to change the font we will assign another spriteAsset to the text

JaXt0r:
Fonts are there.

Assets/GothicVR/Scripts/Player/Menu/MenuDropdownController.cs
JaXt0r: Makes sense. But when you assign the empty stuff and alter it later. It is q reference and all variables will be changed. Even if they shall contain different fonts.
Maybe you can just add a few more comments to make it clear directly at the code snippets.

JaXt0r (approved):
LGTM

PR #192, opened 2023-08-17
persist grabbing between direct and remote grab

Romeeey:
# To test
* [x] Activate remote grabbing
* [x] Remote grab an item
* [ ] Deactive remote grabbing
* [x] Item shall still be in your hand

# Checklist

## Scene

# Testing
* [ ] Tested with PCVR

## Dependencies


JaXt0r (approved):
Impressive. Thanks a lot.

PR #193, opened 2023-08-18
Continuous Collission Detection for VOBItems against World meshes (St…

JaXt0r:
…atic)

# To test
1. Grab some items
2. Throw them... Many times
3. If it will never fall through the world, we're fine


JaXt0r:
Anyone? Please test. It's a 5min thing with 1 line changed. :-D

Tyroonis (approved):
works on my maschine 👍

PR #195, opened 2023-08-18
New main menu

JucanAndreiDaniel:
The scope of this PR is to merge the movement settings and the teleport page into the main menu, and also to have the main menu be available in-game.
# TODO:
* [x] Create settings page containing movement, teleport settings
* [x] Make main menu available in game
* [ ] Create a settings manager to handle saving and changing the settings

# Scene

# Testing
* [ ] Tested with PCVR


Assets/GothicVR/Resources/Prefabs/UI Element Prefabs/MainMenu.prefab
JaXt0r, line 1: I would recommend you rename "UI Element Prefabs" to "UI Elements". The "Prefab" suffix is duplicate as the parent folder is already called "Prefabs".

JaXt0r, line 1: Why do we use a Prefab anyways? It includes all the elements/labels already. How about dynamically create it from phoenix?
image

JaXt0r, line 1: Last but not least: You have the other settings as disabled elements in this Prefab. I think you can safely remove them, right?
image

Assets/GothicVR/Scenes/General.unity
JaXt0r, line 1: Is it intended, that this scene has overrides of the menu prefabs you created before? If not, please apply Overrides to the Prefabs.
image

Assets/GothicVR/Scenes/MainMenu.unity
JaXt0r, line 1: Same here: Can these be Overridden inside the Prefab?
image

Assets/GothicVR/Scripts/Creator/WaynetCreator.cs
```diff
@@ -29,8 +29,10 @@ private void CreateWaypoints(GameObject parent, WorldData world)
 
             foreach (var waypoint in world.waypoints)
             {
-                var wpobject = GameObject.CreatePrimitive(PrimitiveType.Cube);
+                // var wpobject = GameObject.CreatePrimitive(PrimitiveType.Cube);
+                var wpobject = new GameObject();
 
+                wpobject.tag = "PxVob_zCVobSpot";
```
JaXt0r, line 35: Please add this as a Constant inside ConstantsManager.cs

Assets/GothicVR/Scripts/Creator/WaynetCreator.cs
```diff
@@ -29,8 +29,10 @@ private void CreateWaypoints(GameObject parent, WorldData world)
 
             foreach (var waypoint in world.waypoints)
             {
-                var wpobject = GameObject.CreatePrimitive(PrimitiveType.Cube);
+                // var wpobject = GameObject.CreatePrimitive(PrimitiveType.Cube);
+                var wpobject = new GameObject();
```
JaXt0r, line 33: Why do we need this? Because of reasons for spawning?

If we need it for this only, I would recommend we hold xyz coordinates in an array only. Having 5000 additional GameObjects just for getting Vector3 position might be a performance issue for Unity. (As we already have hundreds of GOs)

Assets/GothicVR/Scenes/Bootstrap.unity
```diff
@@ -342,7 +342,7 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   CreateVobs: 1
-  CreateWaypoints: 0
+  CreateWaypoints: 1
```
JaXt0r, line 345: Please revert. I commented insige WayNetCreator.cs how to change behaviour. This debug setting is for visualizing WayPoints only. Not for "having them" ;-)

Assets/GothicVR/Scripts/Player/BasicMovement/ControllerManager.cs
```diff
@@ -65,52 +62,15 @@ public void HideRayCasts()
...
-    public void HideUI()
-    {
-    }
-
-    public void ShowMainMenue()
-    {
-    }
-
-    public void HideMainMenue()
-    {
```
JaXt0r, line 106: Yes. Less code. I like it! 🕺

Assets/GothicVR/Scripts/Player/Menu/MenuDropdownController.cs
```diff
@@ -1,134 +1,60 @@
...
+                "START",
+                "ENTRANCE_SURFACE_OLDMINE",
+                "ENTRANCE_SURFACE_ORCGRAVEYARD",
+                "ENTRANCE_SURFACE_ORCTEMPLE",
+                "ENTRANCE_FREEMINECAMP_FREEMINE",
+                "OCC_CHAPEL_UPSTAIRS",
+                "NC_KDW_CAVE_CENTER",
+                "PSI_TEMPLE_COURT_GURU",
+                "DT_E2_06",
+            };
```
JaXt0r, line 33: You could also think of having this list data created inside scene view. No need to have it here inside code, right?

Assets/GothicVR/Scripts/Player/Menu/MenuManager.cs
```diff
@@ -0,0 +1,87 @@
...
+            SetSettingsValues();
+        }
+
+        public void SetSettingsValues()
+        {
+            var moveSetting = MovementMenu.FindChildRecursively("MoveSpeedSlider").GetComponent<MoveSpeedController>();
+            moveSetting.ChangeMoveSpeed(PlayerPrefs.GetFloat(ConstantsManager.I.moveSpeedPlayerPref));
+
+            var turnSetting = MovementMenu.FindChildRecursively("TurnSettingDropdown").GetComponent<TurnSettingDropdownController>();
+            turnSetting.DropdownItemSelected(PlayerPrefs.GetInt(ConstantsManager.I.turnSettingPlayerPref));
```
JaXt0r, line 27: Can we just assign these values from Scene view? Would be more performant and transparent for coders.

Assets/GothicVR/Scripts/Player/Menu/MenuManager.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        public void PlayFunction()
```
JaXt0r, line 30: Daedalus seems to have actions defined for the menu entries. How about leveraging these and call our methods the same or somethow "map" the information of "which function to call"?
image

Assets/GothicVR/Scripts/Player/Menu/MenuManager.cs
```diff
@@ -0,0 +1,87 @@
...
+            UIMenu.SetActive(true);
+            MainMenu.SetActive(false);
+            SettingsMenu.SetActive(false);
+            TeleportMenu.SetActive(false);
+            MovementMenu.SetActive(false);
+        }
+
+        public void QuitGameFunction()
+        {
+            Application.Quit();
```
JaXt0r, line 84: I really like you already implemented quit. It's a small thing, but if users need to hit Alt+F4 is not good. Thanks.

Assets/GothicVR/Scripts/Player/Menu/TextureMenu.cs
```diff
@@ -11,7 +11,6 @@ public class TextureMenu : MonoBehaviour
         private void Start()
         {
             SetMaterials();
-            FontManager.I.ChangeFont();
```
JaXt0r, line 14: No need any longer? Did you implement Font loading in this branch?

Assets/TextMesh Pro/Resources/TMP Settings.asset
```diff
@@ -15,18 +15,18 @@ MonoBehaviour:
...
   m_GetFontFeaturesAtRuntime: 1
   m_missingGlyphCharacter: 0
   m_ClearDynamicDataOnBuild: 1
-  m_warningsDisabled: 0
-  m_defaultFontAsset: {fileID: 11400000, guid: 8f586378b4e144a9851e7b34d9b748ee, type: 2}
+  m_warningsDisabled: 1
+  m_defaultFontAsset: {fileID: 11400000, guid: ed34b05229166f3469c21b7208879736, type: 2}
   m_defaultFontAssetPath: Fonts & Materials/
-  m_defaultFontSize: 36
+  m_defaultFontSize: 45
```
JaXt0r, line 29: Just FYI if you wanted to have the default font changed.

JaXt0r (changes requested):
I like you removed quite some code. ;-)
Nevertheless, I have a couple of questions.

Once clarified/changed, I'll test your solution.

Assets/GothicVR/Resources/Prefabs/UI Element Prefabs/MainMenu.prefab
JucanAndreiDaniel: Each Menu is a page and we enable/disable them as needed

Assets/GothicVR/Scenes/MainMenu.unity
JucanAndreiDaniel: for the VR player in the Main Menu scene, it's intended

Assets/GothicVR/Scripts/Creator/WaynetCreator.cs
JucanAndreiDaniel: we also need the rotation for teleporting the player

Assets/GothicVR/Scenes/Bootstrap.unity
JucanAndreiDaniel: i use them to be able to teleport the player to more points, since the teleport scrolls teleport the player to waypoints, not freepoints, i added them to the list (as discussed)

Assets/GothicVR/Scripts/Player/Menu/MenuManager.cs
JucanAndreiDaniel: This is meant to load the saved values when loading the VR Player ingame. If you want to change settings in the main menu, we want them to reflect when in-game)

Assets/GothicVR/Scripts/Player/Menu/MenuManager.cs
JucanAndreiDaniel: not using c-menu from daedalus

Assets/GothicVR/Scripts/Player/Menu/TextureMenu.cs
JucanAndreiDaniel: No need to set the font in script if we already use the empty font to force unity to use the spriteAsset containing gothic's font

Assets/GothicVR/Scripts/Player/Menu/MenuManager.cs
Romeeey, line 30: Playfunction is a little generic. How about renaming it to LoadWaypointFromMenu or something.

Assets/GothicVR/Scripts/Manager/ConstantsManager.cs
```diff
@@ -18,8 +18,11 @@ public class ConstantsManager : SingletonBehaviour<ConstantsManager>
 
         //Collection of PlayerPref entries for settings
         public string moveSpeedPlayerPref = "MoveSpeed";
+
         public string turnSettingPlayerPref = "TurnSetting";
 
+        public float moveSpeed = 8f;
+
         public string selectedWorld = "world.zen";
```
Romeeey, line 26: General discussion point: Is constantsmanager the right place to store the loaded world information? Player prefabs are maybe also not right as for options. Maybe it's the same place where player and world parameters are saved eg a json.

Romeeey (changes requested):
In general your implementation looks good.

I would like to request 1 changed before approval:
- Make the waypoints / words better readable (less technical)

When you are on startup / not ingame no locomotion systems seems to be available for the VR player. This could cause a problem with the movement settings. I assume Unity fills in the blanks but maybe we should look at that special case. 

Other than that I have left some comments and ideas for improvement, which you might want to  consider. 

I personally also think that some buttons, labels and forms are too small. But no relevance to this PR. I will try to play around with it and optimize. 

Great work!

Romeeey (approved):

Assets/GothicVR/Scripts/Player/Menu/MenuManager.cs
```diff
@@ -0,0 +1,51 @@
...
+        public GameObject UIMenu;
+
+        void Awake()
+        {
+            SetSettingsValues();
+        }
+
+        public void SetSettingsValues()
+        {
+            var moveSetting = MovementMenu.FindChildRecursively("MoveSpeedSlider").GetComponent<MoveSpeedController>();
```
JaXt0r, line 23: Can you change FindChildRecursively and instead assign the GO in scene view?

Assets/GothicVR/Scripts/Player/Menu/MenuManager.cs
JaXt0r: I think it was called Play in daedalus as well.

JaXt0r (approved):
LGTM. Just one minor comment, but good to go!

PR #198, opened 2023-08-24
109 update to unity 202237f1

Tyroonis:
- Update Unity to 2022.3.7f1
- Update OpenXR
- Update XR Toolkit
- Update TMP
- fix XR Input Asset

**_I didn't copy the input asset and rearrange the input system because @Romeeey  is still working on it.
Would have to take Romeeey in his ticket or we do it later._**

# To test
1. Update Unity
2. Start Projekt
3. Play & check all Functions ;)

# Checklist
- [x] World Loading
- [x] Movement, Teleport, climbing
- [x] Grab
- [x] UI
- [x] Animation correct in the current state
- [x] Audio

## Scene

# Testing
* [x] Tested with PCVR

## Dependencies


JaXt0r (approved):
I didn't check the code changes. :-D

But Standalone works fine.

PR #200, opened 2023-08-29
Separated Vobs into Teleport and NonTeleport entries.

JaXt0r:
# To test
1. Start game (Skip main menu if you like)
2. Check if there are ~100 Collider errors
3. If there are none, we're good

image


JucanAndreiDaniel (approved):
LGTM

PR #201, opened 2023-08-30
VobCulling implemented. (+ GeneralSceneLoaded call moved to UnityEvent)

JaXt0r:
# To test

## Default
1. Change FeatureFlag.VobCulling=false, test performance inside OC
1. Change FeatureFlag.VobCulling=true, test performance inside OC again (Is performance better now? How much?)

## Grabbing
1. Grab an object
2. Move around for more than 100m
3. Drop it down from a height or throw it wide
Expected behaviour: The object is always visible in your hand and when you put it to the ground, it will also stay visible.  
(i.e. in your hand, culling is disabled and once on the floow, it's new position is tracked within Culling)

# Checklist

## Scene

# Testing
* [x] Tested with PCVR


JaXt0r:
I'll implement changeable values for drawDistance and add world mesh + NPCs to it before we can test again.

JaXt0r:
@JucanAndreiDaniel seems we forgot to merge this one. Can you please do a quick test if it still works, so that we can merge asap?

JucanAndreiDaniel:
> @JucanAndreiDaniel seems we forgot to merge this one. Can you please do a quick test if it still works, so that we can merge asap?

We need to fix #205 before being ready to merge

JaXt0r:
> > @JucanAndreiDaniel seems we forgot to merge this one. Can you please do a quick test if it still works, so that we can merge asap?
> 
> We need to fix #205 before being ready to merge

Oh my. This slipped my mind. You're right!

JaXt0r:
I proudly present the second version of Vob Culling. A grabbed object is now fixed and won't be culled any longer. Until it gets dropped and stops falling etc.

JaXt0r:
image
The problem is as you said: The events of "grabbing" are triggered multiple times. I need to handle this.

JaXt0r:
> image The problem is as you said: The events of "grabbing" are triggered multiple times. I need to handle this.

Possible solutions:
1. Check if something's wrong with rigid body setting
2. Add bool flag
3. Or add counter flag
https://discussions.unity.com/t/ontriggerenter-being-called-multiple-times-in-succession/109418/5

JaXt0r:
@JucanAndreiDaniel @Romeeey I stabilized the culling for objects in hand.

Issue was before, that when grabbed an item with >A< and releasing >A<, then OnTriggerExit() was called before OnTriggerEnter() gots recalled ~100ms later.

What to test now: If you grab an item and walk a little bit: Will it disable? -> Shouldn't

JucanAndreiDaniel (approved):

PR #206, opened 2023-09-06
Feature/music tick

JucanAndreiDaniel:
Added a tick system to change the track for the music only once every 5 seconds and also made the buffer for the music bigger to not have the track change so harsh

# To test
1. Music: walk around and listen to the music.
2. FPSpawn: Set SpawnAtSpecificFreePoint at FeatureFlags to get spawned anywhere at an FP (helped JaX to spawn even with XRDeviceSimulator easily)
3. SoundDaytime: Set time to 7:55 and spawn yourself somewhere near the woods. Listen to night crickets etc. and after a few seconds you get birds everywhere

## Scene

# Testing
* [x] Tested with PCVR


Assets/GothicVR/Scripts/Creator/MusicCreator.cs
```diff
@@ -42,6 +42,13 @@ public enum Tags : byte
...
+        private float updateInterval = 5f; // Update music every 5 seconds
+        private float timer = 0f;
+
+        // This multiplier is used to increase the buffer size and reduce the number timess PrepareData is called
+        // also affects the delay of the music, it doesnt sound so harsh when switching
+        private int bufferSizeMultiplier = 16;
+
         void Start()
         {
             if (!FeatureFlags.I.EnableMusic)
```
JaXt0r, line 54: Maybe related to another topic, but: When I start the game, I get the following warning:
image

Can you fix it with this branch as well? Should be a quick win.

Assets/GothicVR/Scripts/Creator/MusicCreator.cs
```diff
@@ -50,6 +57,27 @@ void Start()
             musicSource = backgroundMusic.AddComponent<AudioSource>();
         }
 
+        /// <summary>
+        /// Called once every 0.20 seconds to update the music once the timer is bigger than the updateInterval
+        /// </summary>
+        void FixedUpdate()
```
JaXt0r, line 63: You could also think of using a Coroutine with fixed intervall size.
But to be honest: I don't see why this would break anything inside Unity. ;-)

Assets/GothicVR/Scripts/Creator/MusicCreator.cs
```diff
@@ -50,6 +57,27 @@ void Start()
             musicSource = backgroundMusic.AddComponent<AudioSource>();
         }
 
+        /// <summary>
+        /// Called once every 0.20 seconds to update the music once the timer is bigger than the updateInterval
```
JaXt0r, line 61: Please add an Info "why" you do this. I assume it's because you want to have the music changed at a fixed time frame, right?

Assets/GothicVR/Scripts/Creator/MusicCreator.cs
```diff
@@ -50,6 +57,27 @@ void Start()
...
+        {
+            if (!FeatureFlags.I.EnableMusic)
+                return;
+
+            if (!WorldCreator.I.IsWorldLoaded)
+            {
+                UpdateMusic();
+                return;
+            }
+            timer += Time.fixedDeltaTime;
```
JaXt0r, line 73: Please give the variable a more meaningful name. How about sth. like _musicChangeTimer_ etc.?

Assets/GothicVR/Scripts/Creator/MusicCreator.cs
```diff
@@ -50,6 +57,27 @@ void Start()
...
+            if (timer >= updateInterval)
```
JaXt0r, line 74: Please give the variable a more meaningful name. How about sth. like _changeMusicAfter_ etc.?

Assets/GothicVR/Scripts/Creator/MusicCreator.cs
```diff
@@ -42,6 +42,13 @@ public enum Tags : byte
 
         private static GameObject backgroundMusic;
 
+        private float updateInterval = 5f; // Update music every 5 seconds
+        private float timer = 0f;
+
+        // This multiplier is used to increase the buffer size and reduce the number timess PrepareData is called
+        // also affects the delay of the music, it doesnt sound so harsh when switching
+        private int bufferSizeMultiplier = 16;
```
JaXt0r, line 50: Why 16? Can it be calculated from sth. like updateInterval * x?

Assets/GothicVR/Scripts/Creator/WorldCreator.cs
```diff
@@ -24,8 +24,12 @@ public class WorldCreator : SingletonBehaviour<WorldCreator>
         private GameObject teleportGo;
         private GameObject nonTeleportGo;
 
+        public bool IsWorldLoaded { get; private set; }
```
JaXt0r, line 27: You don't need this. I added UnityEvents for WorldLoading inside this branch: https://github.com/GothicVRProject/GothicVR/pull/201

(As we need to set CullingGroup data once the scene is there). Feel free to check this branch first and reintegrate into your branch second.

JaXt0r: Bonus thing: You can also use the Events for "Unloading" worlds and "stop the MusicUpdate stuff etc.

JaXt0r (changes requested):
A few code questions. I'll test the actual game once clarified.

Assets/GothicVR/Scripts/Creator/MusicCreator.cs
JaXt0r: And when I start main menu, I get this:
image

Assets/GothicVR/Scenes/General.unity
```diff
@@ -142,31 +142,31 @@ PrefabInstance:
...
-      value: -0.78238225
+      value: 0.9849352
       objectReference: {fileID: 0}
     - target: {fileID: 3286136461781483605, guid: 4600f492db92202478d7c53019a519d0, type: 3}
       propertyPath: m_LocalRotation.z
-      value: 0.0063083675
+      value: -0
       objectReference: {fileID: 0}
     - target: {fileID: 3286136461781483605, guid: 4600f492db92202478d7c53019a519d0, type: 3}
       propertyPath: m_LocalEulerAnglesHint.x
```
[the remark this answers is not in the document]
JucanAndreiDaniel: Yes, but they move the default position of the player to the START waypoint instead of being in old camp 
See commit message for more info

Assets/GothicVR/Scripts/Creator/MusicCreator.cs
JucanAndreiDaniel: we can ignore the no audio listeners log, but the multiple listeners could be because you left another scene except bootstrap loaded

Assets/GothicVR/Scripts/Creator/MusicCreator.cs
JucanAndreiDaniel: from manual testing this gives the music a big enough buffer so it sound good when switching tracks

Assets/GothicVR/Scripts/Creator/WorldCreator.cs
JucanAndreiDaniel: Will update to use the UnityEvent once #201 is merged into main

Assets/GothicVR/Scripts/Creator/MusicCreator.cs
JaXt0r: Please don't I would love to have our logs clean as long as possible.
Should be an easy thing to fix, right?

JaXt0r:
I walked around, but music didn't switch. i.e. when entering OC, the OC music didn't start.
Intended?

For debug reasons, you could also add a Debug.Log whenever music is switched.

JaXt0r (changes requested):
I walked around, but music didn't switch. i.e. when entering OC, the OC music didn't start.
Intended?

For debug reasons, you could also add a Debug.Log whenever music is switched.

JaXt0r (changes requested):
The debug logs are helpful. It shows - when entering game on START and moving straight into OC - these themes are called. Two interesting finds:
1. Its often switching between two themes at the same second (in logs)
2. When entering OC, the OC theme isn't played, even the logs tell so

image

JucanAndreiDaniel:
> The debug logs are helpful. It shows - when entering game on START and moving straight into OC - these themes are called. Two interesting finds:
> 
>     1. Its often switching between two themes at the same second (in logs)
> 
>     2. When entering OC, the OC theme isn't played, even the logs tell so
> 
> 
> image

OC and IE are both old camp music (ie is ignored since it's from the same file)
we mostly update the music if a different sgt file is queued

JaXt0r:
Hi @JucanAndreiDaniel I checked OC music switching again.
It seems if we walk into OC and there (within a couple of seconds) switching out again (or to the left, where second OC music boundary starts), then the game switches back to default music from menu or so.

As we discussed some other improvements already: Please let's just have a look into this "OC music change not always switching to OC music" as a last fix. Then we can merge this branch!

JaXt0r:
If I go through the first door at starting area - the start point music will just restart
If I go inside OC, the world music (or so) will start. Moving inside OC between these two overlapping OC elements will restart world music, but not start OC music.

JaXt0r (changes requested):
Bug:
* If I go through the first door at starting area - the start point music will just restart
* If I go inside OC, the world music (or so) will start. Moving inside OC between these two overlapping OC elements will restart world music, but not start OC music.

JaXt0r (approved):
I'm happy. @JucanAndreiDaniel please do one last sanity check with the sounds and then merge.

PR #208, opened 2023-09-12
Feature/settingserrorscreen

Romeeey:
# To test
* [x] Set game settings Gothic 1 path to null or random value
* [x] Start the game
* [x] Check that error message appears with filepicker
* [x] Use file picker to select another wring directory
* [x] Check that error message stays 
* [x] Use filepicker to pick correct directory
* [x] Check that game starts (start menu appears)
* [x] Check that the game settings file have the corresponding values (with escapes \ )
* [x] Restart the game
* [x] Game should start immediately without error message and file picker

The method that checks for the validity of the installation checks for a proper and existing path and for Gothic.exe in \system. We could also check for work and data directories.

# Checklist

## Scene

# Testing
* [ ] Tested with PCVR

## Dependencies


JaXt0r:
@Romeeey how to test? Please add the test scenario.

Assets/StandaloneFileBrowser/Plugins/System.Windows.Forms.dll
JaXt0r, line 1: Will this work on Linux? I.e. if the path to G1 is wrong on Android it wants to open this pop-up too, right? How to prevent?

Assets/StandaloneFileBrowser/Plugins/Linux/x86_64/libStandaloneFileBrowser.so.meta
JaXt0r, line 1: Linux .so aren't needed. The plugin shall o my work on Windows, right?

Assets/StandaloneFileBrowser.meta
JaXt0r, line 1: Please add an information inside our doc/Dev readme about the plugin, where we have it from and why we use it.

Assets/StandaloneFileBrowser/StandaloneFileBrowserWindows.cs.meta
JaXt0r, line 1: Did you pull the asset package from Unity asset store or via GitHub directly? We could reference it inside package.json of our project. Then it will load it dynamically and we don't need to store any file here.

I did this before. Please ping me.

JaXt0r (changes requested):

Assets/GothicVR/Scripts/Manager/Settings/SettingsManager.cs
```diff
@@ -67,13 +78,16 @@ private string GetRootPath()
...
-            if (!Directory.Exists(GameSettings.GothicIPath))
-                throw new ArgumentException(
-                    $"GothicI installation path wasn't found at >{GameSettings.GothicIPath}<." +
-                    $"Please put in the right absolute path to your local installation.");
-        }
+            bool isValid = false;
+            if (Directory.Exists(GameSettings.GothicIPath))
+            {
+                var pathGothicExe = GameSettings.GothicIPath + "\\system\\Gothic.exe";
+                isValid = File.Exists(pathGothicExe);
```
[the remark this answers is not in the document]
Romeeey: I think I return a bool. Therefore I used the variable.

Assets/GothicVR/Scripts/Bootstrap/PhoenixBootstrapper.cs
```diff
@@ -1,31 +1,33 @@
...
     {
         private bool _loaded = false;
+        private GameObject installationFilePicker;
 
         private void Start()
         {
             PxLogging.pxLoggerSet(PxLoggerCallback);
+
+            //Initialize and switch off FilepickerUI
+            installationFilePicker = GameObject.Find("InstallationFilePicker");
```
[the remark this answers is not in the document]
Romeeey: That's the alternative, yeah. Normally I do it that way. The disadvantage is that the UI looks more complicated although its not necessary (the canvas will always be the same) and sometimes these references are lost. 

Maybe the find option is better. But I don't know about performance.

Assets/GothicVR/Scripts/UI/OpenSettingsFilePicker.cs
```diff
@@ -0,0 +1,26 @@
...
+        public static void OpenGothicInstallationFilePicker()
+        {
+            var paths = StandaloneFileBrowser.OpenFolderPanel("Select Gothic installation","", false);
+
+            if (paths != null && paths.Length > 0)
+            {
+                GameSettings gameSettings = SettingsManager.I.LoadGameSettings();
+                gameSettings.GothicIPath = paths[0];
+                SettingsManager.I.SaveGameSettings(gameSettings);
+                if (SettingsManager.I.CheckIfGothic1InstallationExists())
```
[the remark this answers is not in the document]
Romeeey: The file picker closes, the code will try to start the game (check the validity of the installation via existing path and gothic.exe), fail and the error message remains. A use has the opportunity to use the file picker again.

Assets/StandaloneFileBrowser/Plugins/System.Windows.Forms.dll
Romeeey: Does a Gothic installation run on Linux? The file picker I used before was a little more versatile. This one that Thomas suggested is a native Windows file picker. 

In one method I check that it shouldn't do it on Android. But you are right, we need to check for the entire implementation, that its only PCVR. There are multiple options for Windows when checking the platform. Do you know which one the correct one is?

Assets/StandaloneFileBrowser/Plugins/Linux/x86_64/libStandaloneFileBrowser.so.meta
Romeeey: Yes

Assets/StandaloneFileBrowser.meta
Romeeey: Ok. Will do. 

Its a little bit weird because the plugin goes directly in the root of Unity. Moving it to plugins causes errors.

Assets/StandaloneFileBrowser/StandaloneFileBrowserWindows.cs.meta
Romeeey: Ok. I went to the github page of the project, but funnily enough got a unity package.

Assets/GothicVR/Scripts/Bootstrap/PhoenixBootstrapper.cs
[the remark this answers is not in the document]
Tyroonis: @Romeeey 
Attaching it directly to a GO is more performant.
GameObject.Find searches all GOs and that costs CPU.

Assets/GothicVR/Scripts/Manager/Settings/SettingsManager.cs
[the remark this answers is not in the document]
Tyroonis: But what if I want to take only one folder with DATA and _work? That would be good if that also goes.
Because I have modded my original game e.g.

Does not it make more sense to see if the 2 necessary folders are there?

Assets/GothicVR/Scripts/Manager/Settings/SettingsManager.cs
```diff
@@ -43,9 +54,9 @@ private void LoadGameSettings()
             {
                 var devJson = File.ReadAllText(settingsDevFilePath);
                 JsonUtility.FromJsonOverwrite(devJson, GameSettings);
+                return GameSettings;
```
JaXt0r, line 57: Oh. This one prevented me from using the GameSettings.dev.json file.

JaXt0r:
@Romeeey I updated a few things. Please check my comments and changes. If fine for you, I'm also good to approve the PR.

Assets/GothicVR/Scripts/Bootstrap/PhoenixBootstrapper.cs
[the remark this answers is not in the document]
Romeeey: Ok. Change done.

Assets/GothicVR/Scripts/Bootstrap/PhoenixBootstrapper.cs
```diff
@@ -1,31 +1,33 @@
...
+using UnityEngine;
 using Debug = UnityEngine.Debug;
 
 namespace GVR.Bootstrap
 {
     public class PhoenixBootstrapper : SingletonBehaviour<PhoenixBootstrapper>
     {
         private bool _loaded = false;
+        public GameObject configurationMessage;
+        public GameObject filePickerButton;
```
JaXt0r, line 23: These two aren't set in Bootstrap scene. I get a nullpointer exception.
image

Assets/GothicVR/Scripts/UI/OpenSettingsFilePicker.cs
```diff
@@ -0,0 +1,28 @@
...
+            if (paths == null || paths.Length == 0)
+                return;
+            
+            GameSettings gameSettings = SettingsManager.I.LoadGameSettings();
+            gameSettings.GothicIPath = paths[0];
+            SettingsManager.I.SaveGameSettings(gameSettings);
+            
+            if (SettingsManager.I.CheckIfGothic1InstallationExists())
+            {
+                PhoenixBootstrapper.I.installationFilePicker.SetActive(false);
```
JaXt0r, line 23: This property is renamed. I assume we need to disable both: The button and the message?
image

Assets/GothicVR/Scripts/Bootstrap/PhoenixBootstrapper.cs
```diff
@@ -1,31 +1,33 @@
...
 
         private void Start()
         {
             PxLogging.pxLoggerSet(PxLoggerCallback);
+
+            // Just in case we forgot to disable it in scene view. ;-)
+            configurationMessage.SetActive(false);
```
JaXt0r, line 30: I think we also should disable the button, right? Just in case we forgot to disable it inside the scene.

JaXt0r (changes requested):
Two bugs.

Assets/StandaloneFileBrowser/StandaloneFileBrowserWindows.cs.meta
JaXt0r: Yeah. I checked it. It isn't possible to import it via Package manager, as it won't comply with file needs (package.json or so is missing). i.e. you did it right in importing it.

Assets/StandaloneFileBrowser.meta
JaXt0r: I moved it and added a README about the repo where we got it from.

Assets/StandaloneFileBrowser/Plugins/Linux/x86_64/libStandaloneFileBrowser.so.meta
JaXt0r: We can just leave the .so for now. It's a few kB only. No harm. :-)

Assets/GothicVR/Scripts/Manager/Settings/SettingsManager.cs
JaXt0r: Fixed. I changed it.

Assets/StandaloneFileBrowser/Plugins/System.Windows.Forms.dll
JaXt0r: You fixed it in your last commit: If Standalone, then only message, no filepicker.

Assets/GothicVR/Scripts/Manager/Settings/SettingsManager.cs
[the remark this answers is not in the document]
JaXt0r: Changed it as suggested.

Assets/GothicVR/Scripts/Bootstrap/PhoenixBootstrapper.cs
JaXt0r:
```suggestion
            configurationMessage.SetActive(false);
            filePickerButton.SetActive(false);
```

Assets/GothicVR/Scripts/Bootstrap/PhoenixBootstrapper.cs
```diff
@@ -35,17 +37,34 @@ private void Update()
...
+                BootGothicVR(g1Dir);
+            }
+            else
+            {
+                //Show the startup config message, show filepicker for PCVR but not for Android standalone
+                configurationMessage.SetActive(true);
+                if (Application.platform == RuntimePlatform.Android)
+                    filePickerButton.SetActive(false);
+                else
+                    filePickerButton.SetActive(true);
```
JaXt0r, line 53:
```suggestion
                // Activate filePicker if we're not on Android
                filePickerButton.SetActive(Application.platform != RuntimePlatform.Android);
```

JaXt0r (changes requested):
Small changes/questions only.

JaXt0r (changes requested):
Unfortunately I get a pink "missing material" errors when starting the game with a broken installation.

image

JaXt0r:
And also when in Editor mode, I can't click the button. (I can't really remember if it worked this way before. This would be ok, but the pink build needs to be fixed.)

image

Assets/GothicVR/Scripts/UI/OpenSettingsFilePicker.cs
Romeeey: No. The basic idea behind the last little implementation is that the text always show. But the button is only visible in PCVR. This gives the PCVR user the possibility to configure the path via the file picker. An Android user only gets the information that the files are missing. We might wanna show a different more detailed message for Android.

Romeeey:
> Unfortunately I get a pink "missing material" errors when starting the game with a broken installation path.
> 
> image

Added the UI / Default shader to the project. (this will also fix minor bugs of Andrei).
Merged with main.
Retested with new build.

JaXt0r (approved):
Works fine now!
--> Message shows only when path is wrong
--> Build&Run works and the texture is there now

Didn't test it on Mobile VR, but I assume you did it.

PR #211, opened 2023-09-17
Feature/207 add xrdevicesimulator

JaXt0r:
# To test
1. Set Feature Flag: _Use XR Device Simulator_
2. Try to navigate through the main menu (T + Left Mouse)
3. Try to navigate through the world (Z + W = Teleport; T + G= Grab; T + WASD = Movement)
4. If you deactivate the setting, no WASD controls should work any longer

Hints:
* The simulator is activated for Editor mode only!
* The first 5 files or so need to be checked. The others are the neccessary sample files from OpenXR
* Documentation mentions the controls at /Docu/development/developer-hints.md

# Bugs
I can't test with normal controls/main branch if the bugs are related to this new feature or were there before. Please check both with main branch and your HMD.

## Bug 1
I can' switch to anothr world. The loading screen stops without exception. But I don't think it's related to the new XRDeviceSimulator. I can't test my assumption these days, can you please do for me.

## Bug 2
when I start game normally, select "New Game" and then activate teleportation (and release it), I get the exceptions shown on screenshot every time. (Hint: If we select FeatureFlag.SkipMainMenu, we don't get it)

I have two ideas:
It is because we recreate/reuse the VRPlayer with some destroyed references (as the error mentions)
It is because of my XRDeviceSimulator doing some magic

I can't check it with a real HMD for the next days and therefore wanted to ask you for your feedback, if we have this exception on main branch and with normal HMD usage as well.
image

# Checklist
## Scene

# Testing


JucanAndreiDaniel (approved):
LGTM


Bug 1 will be fixed in another PR as it was caused by #200 
Bug 2 is not related to this PR

PR #212, opened 2023-09-18
bugfix/reset the list with the nonTeleport GameObjects

JucanAndreiDaniel:
# To test
1. Go to another world (e.g. old mine, free mine)

# Testing
* [x] Tested with PCVR


JaXt0r (approved):
Definitely a missed line. But why does it help to load new worlds?

PR #216, opened 2023-10-03
Feature/zs idle animations - initial logic and first walking/sleeping

JaXt0r:
# To test
1. Check code
1. Check developer documentation (/Docs/...) for some hints about architecture.
2. Check if the game compiles
3. If Npcs and Vobs are activated: Check if there is one NPC who is doing idle animations. Whatever it is. It just needs to loop for multiple idle animations. ;-)

There are more idle animations already existing, but it's difficult to set them up for testing. For now it would be great if code looks fine for you and makes sense. After merging, I will work on Routines, so that Bloodwyn will walk through OC and other NPCs will start eating stuff and sleeping at night.

# Checklist

## Scene

# Testing
* [x] Tested with PCVR

## Dependencies


JaXt0r:
@JucanAndreiDaniel I merged main into this one. Took me 2h. :-D

It would be great if you can briefly check this one, that I never ever need to merge it again. :-D

Tyroonis (approved):

PR #223, opened 2023-10-11
Feature/separate world meshes

JaXt0r:
This branch is for a later use. It includes WorldMesh culling and separated submeshes (6k for world.zen). We figured out, that performance is worse than with it. But the two implementations might come in handy in the future, when we work on an improved world loading from and with Phoenix. This PR is just to save it inside main and see if we will reuse it in the future (without rebuilding it from scratch).

image

# To test

- [ ] Check code for bugs/code smells
- [ ] Check if FeatureFlags disable the two Experimental flags
- [ ] You could potentially activate both Experimental FeatureFlags but performance will be worse than before.


Tyroonis:
image

Looks good. 
Feature flags works

Tyroonis (approved):

PR #226, opened 2023-10-12
Alpha to coverage shader for alpha clipped objects

ThomasMountainborn:
Greatly reduce aliasing on alpha clipped textures. Separate shaders for opaque and alpha clipped materials. Manually generate mips for textures with transparency.

### DIFF
--- a/Assets/GothicVR/Scripts/Caches/AssetCache.cs
+++ b/Assets/GothicVR/Scripts/Caches/AssetCache.cs
@@ -71,16 +71,30 @@ namespace GVR.Caches
             var format = pxTexture.format.AsUnityTextureFormat();
 
             Texture2D texture = null;
-            if (pxTexture.format == PxTexture.Format.tex_B8G8R8A8)
+            if (pxTexture.mipmapCount == 1)
             {
-                // Let Unity generate mips for textures with alpha, as the game doesn't provide them.
-                texture = new Texture2D((int)pxTexture.width, (int)pxTexture.height, format, true);
-                texture.SetPixelData(pxTexture.mipmaps[0].mipmap, 0);
-                texture.Apply(true, true);
+                // Let Unity generate mips if not provided.
+                if (format == TextureFormat.DXT1)
+                {
+                    // Unity doesn't want to create mips for DXT1 textures. Recreate them as RGB24.
+                    Texture2D dxtTexture = new Texture2D((int)pxTexture.width, (int)pxTexture.height, format, false);
+                    dxtTexture.SetPixelData(pxTexture.mipmaps[0].mipmap, 0);
+                    dxtTexture.Apply(false);
+                    texture = new Texture2D((int)pxTexture.width, (int)pxTexture.height, TextureFormat.RGB24, true);
+                    texture.SetPixels(dxtTexture.GetPixels());
+                    texture.Apply(true, true);
+                    Destroy(dxtTexture);
+                }
+                else
+                {
+                    texture = new Texture2D((int)pxTexture.width, (int)pxTexture.height, format, true);
+                    texture.SetPixelData(pxTexture.mipmaps[0].mipmap, 0);
+                    texture.Apply(true, true);
+                }
             }
             else
             {
-                // Use Gothic's mips for opaque textures. We could also let Unity generate them here, though.
+                // Use Gothic's mips if provided. We could also let Unity generate them here, though.
                 texture = new Texture2D((int)pxTexture.width, (int)pxTexture.height, format, (int)pxTexture.mipmapCount, false);
                 for (int i = 0; i < pxTexture.mipmapCount; i++)
                 {
--- a/Assets/GothicVR/Scripts/Caches/AssetCache.cs
+++ b/Assets/GothicVR/Scripts/Caches/AssetCache.cs
@@ -69,14 +69,26 @@ namespace GVR.Caches
             }
 
             var format = pxTexture.format.AsUnityTextureFormat();
-            var texture = new Texture2D((int)pxTexture.width, (int)pxTexture.height, format, (int)pxTexture.mipmapCount, false);
+
+            Texture2D texture = null;
+            if (pxTexture.format == PxTexture.Format.tex_B8G8R8A8)
+            {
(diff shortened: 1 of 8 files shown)


Assets/GothicVR/Scripts/Caches/AssetCache.cs
```diff
@@ -69,14 +69,26 @@ public Texture2D TryGetTexture(string key)
...
-            texture.name = key;
-
-            for (var i = 0u; i < pxTexture.mipmapCount; i++)
-                texture.SetPixelData(pxTexture.mipmaps[i].mipmap, (int)i);
 
-            texture.Apply();
+            Texture2D texture = null;
+            if (pxTexture.format == PxTexture.Format.tex_B8G8R8A8)
+            {
+                texture = new Texture2D((int)pxTexture.width, (int)pxTexture.height, format, true);
```
JaXt0r, line 76: Why not using MipMaps for this format? Please add a comment about it.

Assets/GothicVR/Scripts/Creator/Meshes/AbstractMeshCreator.cs
```diff
@@ -492,14 +499,18 @@ protected virtual Texture2D GetTexture(string name)
...
-            var standardShader = Shader.Find(defaultShader);
-            var material = new Material(standardShader);
-
-            // Enable clipping of alpha values.
-            material.EnableKeyword("_ALPHATEST_ON");
-
+            var shader = Shader.Find(defaultShader);
+            if (isAlphaTest)
+            {
+                shader = Shader.Find("Unlit/Unlit-AlphaToCoverage");
```
JaXt0r, line 507: Nie one. Please put this string as const on top of the class.

JaXt0r, line 507: As we don't reference shaders inside scene, we also need this shader to be added to Project Settings --> Graphics --> Always Includes Shaders

Assets/GothicVR/Scripts/Creator/Meshes/AbstractMeshCreator.cs
```diff
@@ -492,14 +499,18 @@ protected virtual Texture2D GetTexture(string name)
...
+            }
+            var material = new Material(shader);
+            if (isAlphaTest)
+            {
+                material.renderQueue = 2450;
```
JaXt0r, line 512: Why this value? Please put this as constant on top (#MagicNumber) and add a comment why this value.

Assets/GothicVR/Settings/URP-Balanced.asset
```diff
@@ -23,9 +23,9 @@ MonoBehaviour:
   m_RequireOpaqueTexture: 0
   m_OpaqueDownsampling: 1
   m_SupportsTerrainHoles: 1
-  m_SupportsHDR: 1
+  m_SupportsHDR: 0
   m_HDRColorBufferPrecision: 0
-  m_MSAA: 1
+  m_MSAA: 4
```
JaXt0r, line 28: Oh my. Seems it got reset after Frank changed it last time. Thanks.

Assets/GothicVR/Shaders/Unlit-AlphaToCoverage.shader
```diff
@@ -0,0 +1,86 @@
+Shader "Unlit/Unlit-AlphaToCoverage"
```
JaXt0r, line 1: Hm. Maybe this shader will work without setting it to "Always Included Shaders".
I thought we need to put it inside _Resources_ folder to be found at runtime. Will it work for Standalone builds?

JaXt0r (changes requested):
A few questions and Standalone test only.
I will check on my HMD afterwards.

Assets/GothicVR/Shaders/Unlit-AlphaToCoverage.shader
ThomasMountainborn: It is included in always included shaders. 😘 https://github.com/GothicVRProject/GothicVR/pull/226/commits/31ef83133933de5f6570e9e58f22b0b68d0ca0b4#diff-d6b5b7d6a42d4e04b725021f38ab5029472e26932c63c9ba1a03a3bee95a9a1d

Assets/GothicVR/Scripts/Caches/AssetCache.cs
```diff
@@ -69,14 +69,28 @@ public Texture2D TryGetTexture(string key)
...
-
-            for (var i = 0u; i < pxTexture.mipmapCount; i++)
-                texture.SetPixelData(pxTexture.mipmaps[i].mipmap, (int)i);
 
-            texture.Apply();
+            Texture2D texture = null;
+            if (pxTexture.format == PxTexture.Format.tex_B8G8R8A8)
+            {
+                // Let Unity generate mips for textures with alpha, as the game doesn't provide them.
+                texture = new Texture2D((int)pxTexture.width, (int)pxTexture.height, format, true);
```
JaXt0r, line 77: This doesn't seem to be true. There are objects which have mipmaps even when using B8... texture format.
Wouldn't it be better to check if mipmapCount == 1 and then let Unity create them?
image


Reason for this is: If the texture is not in a Unity supported format (dxt1, dxt5) it will be converted to B8...

image

image

JaXt0r (changes requested):
Last question about manual mipmap creation.

Assets/GothicVR/Scripts/Caches/AssetCache.cs
ThomasMountainborn: Honestly Unity's mips are going to be better anyway probably, but sure.

JaXt0r: I understand. It goes into the direction "classical G1 experience" vs improved one. I count better mipmaps to the latter. From my point we're still in the direction of the classical experience. If it won't cost performance or other awkward stuff. ;-)

Cool. After this one, we're good to go. Thank you.

JaXt0r (changes requested):

JaXt0r (approved):
LGTM

PR #234, opened 2023-10-18
Re-added collission for Vobs.

JaXt0r:
# To test
1. Check if you fall through the gate inside OC inner ring.
2. Check if you can not walk through barrels
3. Check if you can still walk onto beds


JaXt0r:
Included in other branch. My bad.

PR #236, opened 2023-10-18
New occlusion culling data and frustum culling fix

ThomasMountainborn:
New occlusion culling bake with alpha clipped meshes not set as occluders. Fix culling group spheres being centered around mesh origin instead of mesh center.

# To test
If incorrect culling still occurs, and standalone VR is still running at 72Hz everywhere.

### DIFF
--- a/Assets/GothicVR/Scripts/Manager/Culling/VobMeshCullingManager.cs
+++ b/Assets/GothicVR/Scripts/Manager/Culling/VobMeshCullingManager.cs
@@ -15,9 +15,6 @@ namespace GVR.Manager.Culling
     /// </summary>
     public class VobMeshCullingManager : SingletonBehaviour<VobMeshCullingManager>
     {
-        [SerializeField]
-        protected bool drawGizmos;
-
         // Stored for resetting after world switch
         private CullingGroup vobCullingGroupSmall;
         private CullingGroup vobCullingGroupMedium;
@@ -58,9 +55,12 @@ namespace GVR.Manager.Culling
             StartCoroutine(StopVobTrackingBasedOnVelocity());
         }
 
+        /// <summary>
+        /// This method will only be called within EditorMode. It's tested to not being executed within Standalone mode.
+        /// </summary>
         private void OnDrawGizmos()
         {
-            if (!drawGizmos)
+            if (!Application.isPlaying || !FeatureFlags.I.drawVobCullingGizmos)
             {
                 return;
             }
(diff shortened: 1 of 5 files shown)


Assets/GothicVR/Scripts/Manager/Culling/VobMeshCullingManager.cs
```diff
@@ -15,11 +15,14 @@ namespace GVR.Manager.Culling
     /// </summary>
     public class VobMeshCullingManager : SingletonBehaviour<VobMeshCullingManager>
     {
+        [SerializeField]
+        protected bool drawGizmos;
```
JaXt0r, line 19: We put this kind of setting into FeatureFlags.
There should be a section of _Developer_ flags.

Please put it there and name it like _debugDrawCullingGizmosOnVobs_ to have it centrally available.

JaXt0r (changes requested):
One small request.

Just to double check: You didn't need to create cullingData for worlds new? Currently I see the fix is only 1 line of code (which is great ;-) )

JaXt0r:
Ok. I figured out, that you changed Culling for world.zen.
Please execute it for the other worlds as well.

ThomasMountainborn:
Added occlusion bakes for orc temple and graveyard. The other worlds don't have alpha clipped geometry.

JaXt0r (changes requested):
Solution works. No "flickering" of Vobs on screen (tested with PCVR standalone build).
Two requests based on documentation and code, then we're done.

Tyroonis:
Works also an Pico.

Assets/GothicVR/Scripts/Manager/Culling/VobMeshCullingManager.cs
```diff
@@ -15,11 +15,14 @@ namespace GVR.Manager.Culling
     /// </summary>
     public class VobMeshCullingManager : SingletonBehaviour<VobMeshCullingManager>
```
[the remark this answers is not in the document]
JaXt0r: We discussed. Nothing to document as it will be added with another branch.

Assets/GothicVR/Scripts/Manager/Culling/VobMeshCullingManager.cs
JaXt0r: Last point on the PR. Then I merge. I can also do the change for you. Just tell me what's faster. @ThomasMountainborn

ThomasMountainborn: ### Can you do it? Thanks!

JaXt0r (approved):

JaXt0r:
I just merge it as it's been tested and worked on from two people.

PR #237, opened 2023-10-19
Bugfix/ladder grabbing two handed

JaXt0r:
# To test
1. Spawn into world
2. Grab the first ladder and climb upwards downwards - one handed left
3. Same one handed right
4. Same two handed
5. Check if there are any warnings, errors, etc. on screen
6. Check if your right hand can also climb if you work two handed

Not part of this fix:
* putting the settings into a ladder Prefab. Will be another ticket soon. (Whole vob types need to be put into more fine grained prefabs)
* There is no snap when reaching top. I.e. makes reaching a higher floor quite hard. - should create separate ticket for that.
* teleport onto ladder top isn't checked. I think it's kind of broken, but should be handled in another PR afterwards

### DIFF
--- a/Assets/GothicVR/Scripts/Creator/Meshes/VobMeshCreator.cs
+++ b/Assets/GothicVR/Scripts/Creator/Meshes/VobMeshCreator.cs
@@ -37,7 +37,7 @@ namespace GVR.Creator.Meshes
 
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


JucanAndreiDaniel (approved):
Tested, works fine, and no errors thrown

PR #251, opened 2023-11-03
Feature/vob particlefx

JaXt0r:
This PR is intended to merge first Particle implementation. It's definitely WIP (e.g. fire) but the creation flow is fine. I want to merge it and will work on it at a later stage again.

# To test
1. Activate FeatureFlag: _enableVobParticles_ (under Experimental)
2. Check if you see properly flying leaves inside woods
4. Check if you see broken fire particles (expected)

# Checklist

## Scene

# Testing
* [x] Tested with PCVR

## Dependencies


Assets/GothicVR/Scripts/Creator/VobCreator.cs
```diff
@@ -591,23 +599,155 @@ private static GameObject CreateDecal(PxVobData vob)
...
+                    forceModule.x = gravityX;
+                    forceModule.y = gravityY;
+                    forceModule.z = gravityZ;
+                }
+            }
+
+            // Renderer module
+            {
+                var rendererModule = go.GetComponent<ParticleSystemRenderer>();
+                TextureManager.I.SetTexture(pfx.visName, rendererModule.material);
```
JucanAndreiDaniel, line 685: Default selected shader for the material is URP/Lit which takes a ton of time to compile (even if it's one time only)
Maybe we can change it do Simple Lit or another shader which doesn'y take that much to compile
```suggestion
                var standardShader = Shader.Find("Universal Render Pipeline/Simple Lit");
                var material = new Material(standardShader);
                rendererModule.material = material;
                TextureManager.I.SetTexture(pfx.visName, rendererModule.material);
```

JucanAndreiDaniel (approved):
Seems fine for a first approach for particles, but need to keep in mind that the emitter gameobject is culled, so the particles might disappear when walking too far away from the emitter

JaXt0r:
Nice teamwork @JucanAndreiDaniel . It's been a pleasure as alway. :-)

PR #255, opened 2023-11-04
FeatureFlag tool implemented and documented and in pipeline

JaXt0r:
…ring release creation.

# To test
1. ...
2. ...
3. ...

# Checklist

## Scene

# Testing
* [ ] Tested with PCVR

## Dependencies


JaXt0r:
Tested and worked on by Tyroonis and JaXt0r.
Will merge it now to test last feature: ReleaseBuild handling. As it's only triggered from main.

PR #257, opened 2023-11-05
fix directorySearch

Tyroonis:
# To test
1. Start GVR on MVR without Files --> Error
2. Start GVR on MVR with Files --> GVR starts
3. Start GVR on PCV with wrong Path --> Filepicker
4. Start GVR on PCVR with correct Path --> GVR starts

# Checklist

## Scene

# Testing
* [ ] Tested with PCVR
* [x] Tested with Pico
* [ ] Tested with Quest

## Dependencies

### DIFF
--- a/Assets/GothicVR/Scripts/Manager/Settings/SettingsManager.cs
+++ b/Assets/GothicVR/Scripts/Manager/Settings/SettingsManager.cs
@@ -76,9 +76,11 @@ namespace GVR.Manager.Settings
 
         public static bool CheckIfGothic1InstallationExists()
         {
-            if (Directory.Exists(GameSettings.GothicIPath))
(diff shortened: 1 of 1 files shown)


Tyroonis:
@JaXt0r @Romeeey 
android cannot handle " Directory.Exists".
It always comes back false.
I have modified the check a little.

JaXt0r (changes requested):
Unfortunately not yet.
On PCVR it always returns the path even if it's nonsense. (Path.GetFullPath() is no real "checker" but more a string concatenation)

Can you please check again why Directory.Exists didn't work on Android? Maybe the path was broken and you need something like file:// in front?

Idea: Use a hard coded string to test on MVR. I assume it's the path structure, not the call itself.

JucanAndreiDaniel:
Proper fix implemented in #258

PR #258, opened 2023-11-07
fix directory checks for instalation folders

JucanAndreiDaniel:
# Testing
* [x] Tested with PCVR


Tyroonis (approved):
works on Pico and PCVR

PR #259, opened 2023-11-08
Particles small update

JucanAndreiDaniel:
Color over lifetime, proper offset used and size over lifetime implemented

### DIFF
--- a/Assets/GothicVR/Scripts/Manager/Culling/VobMeshCullingManager.cs
+++ b/Assets/GothicVR/Scripts/Manager/Culling/VobMeshCullingManager.cs
@@ -231,16 +231,24 @@ namespace GVR.Manager.Culling
         {
             var transf = go.transform;
 
-            if (transf.TryGetComponent<MeshFilter>(out var mesh0)) // Lookup: /
-                return mesh0.mesh;
-            else if (transf.GetChild(0).TryGetComponent<MeshFilter>(out var mesh1)) // Lookup: /BIP 01
-                return mesh1.mesh;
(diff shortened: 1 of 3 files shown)


Assets/GothicVR/Scripts/Manager/Culling/VobMeshCullingManager.cs
```diff
@@ -231,16 +231,24 @@ private Mesh GetMesh(GameObject go)
...
-            if (transf.TryGetComponent<MeshFilter>(out var mesh0)) // Lookup: /
-                return mesh0.mesh;
-            else if (transf.GetChild(0).TryGetComponent<MeshFilter>(out var mesh1)) // Lookup: /BIP 01
-                return mesh1.mesh;
-            else if (transf.GetChild(0).GetChild(0).TryGetComponent<MeshFilter>(out var mesh2)) // Lookup: /BIP 01/...
-                return mesh2.mesh;
-            else if (transf.childCount > 1 && transf.GetChild(1).TryGetComponent<MeshFilter>(out var mesh3)) // Lookup: /ZM_0
-                return mesh3.mesh;
-            else
+            try
```
JaXt0r, line 234: Why a try catch? I already added TryGet() methods.  There shouldn't be an error. Just plain return null at max.

JaXt0r (changes requested):
One question. Rest looks good.

Assets/GothicVR/Scripts/Manager/Culling/VobMeshCullingManager.cs
JucanAndreiDaniel: It seems to throw an error at mesh2 about transform out of bounds, so I just put another try catch to ignore it, this is caused by particles

JaXt0r (approved):

PR #261, opened 2023-11-12
Feature/sound volume sliders

JucanAndreiDaniel:
# To test
1. Test volume in main menu
2. Load world and test if volume is properly set
3. Restart the game and check if sound volume is properly set

# Testing
* [x] Tested with PCVR

### DIFF
--- a/Assets/GothicVR/Scenes/Bootstrap.unity
+++ b/Assets/GothicVR/Scenes/Bootstrap.unity
@@ -1236,6 +1236,7 @@ GameObject:
   serializedVersion: 6
   m_Component:
   - component: {fileID: 1283201086}
+  - component: {fileID: 1283201087}
   m_Layer: 0
   m_Name: BackgroundMusic
   m_TagString: Untagged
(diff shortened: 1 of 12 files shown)


Assets/GothicVR/Scenes/Bootstrap.unity
```diff
@@ -1258,6 +1259,102 @@ Transform:
   m_Children: []
   m_Father: {fileID: 1103036566}
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
+--- !u!82 &1283201087
+AudioSource:
```
JaXt0r, line 1263: Why do we need an AudioSource on Bootstrap scene? Wouldn't it be better suited inside Menu.unity or General.unity?

Assets/GothicVR/Scripts/UI/AudioMixerHandler.cs
```diff
@@ -0,0 +1,25 @@
...
+        void Awake()
+        {
+            float oldVolume = PlayerPrefs.GetFloat(volumePlayerPrefName, 1f);
+            audioVolumeSlider.value = oldVolume;
+        }
+
+        public void SliderUpdate(float value)
+        {
+            PlayerPrefs.SetFloat(volumePlayerPrefName, value);
+            audioMixer.audioMixer.SetFloat(volumePlayerPrefName, Mathf.Log10(value) * 20);
```
JaXt0r, line 22: Why log(value) * 20?
Can a slider value like 0-100 be sufficient?

JaXt0r (changes requested):

Assets/GothicVR/Scenes/Bootstrap.unity
JucanAndreiDaniel: Before we already had the BackgroundMusic gameobject being available in the Bootstrap scene, but the audio source was created in script in MusicManager.
I just skipped the step in the script to create the Audio source, just so i can already set the output mixer, letting me control the volume perceived by the audio listener without directly controlling the volume of the audio source

Assets/GothicVR/Scripts/UI/AudioMixerHandler.cs
JucanAndreiDaniel: Volume and loudness are not the same, volume can be linear but loudness is logarithmic
https://www.msdmanuals.com/home/multimedia/table/measurement-of-loudness

JaXt0r (commented):
Thanks for clarifications. Code looks good now. Will test the implementation soon.

PR #262, opened 2023-11-13
Feature/improve settings

Romeeey:
# To test
1. Start the game
2. Change movement setting for turntype from snap to continuous turn
3. Check that continuous turn is on
4. Go to different map (eg old mine)
5. Check that continuous turn is still on
6. Go back to surface map and check that it's still continuous
7. Change the movement turn type to snap back in the editor (playercontroller gameobject)
8. Check that the turn type changed ingame

# Comments to the code
- A split between business logic and UI was done.
- The business logic lies within the PlayerSettingsManager
- The UI ingame and feature flags reference the PlayerSettingsManger
- The turntype exists as 2 variables due to the fact that there is the ingame UI, the editor UI and the real static variable
- Whenever the playercontroller is created (after destroying the old scene and creating the new scene) the settings are loaded from playerprefs
- OnValidate is restricted to UI changes in the editor (otherwise it would overwrite the playerprefs because it is also called when the script is created and is conflicting with the awake of the playercontroller)

# Checklist

## Scene

# Testing
* [ ] Tested with PCVR

## Dependencies


JaXt0r:
@JucanAndreiDaniel @Romeeey 
When shall we test?

JaXt0r:
@JucanAndreiDaniel @Romeeey still valid branch? If so, please advice next steps. If not, feel free to remove it.

PR #265, opened 2023-11-14
Dynamic Interactables

BoroBongo:
https://discord.com/channels/1000261958668406874/1173448970417213510 

# To test
1. ...
2. ...
3. ...

# Checklist

## Scene

# Testing
* [ ] Tested with PCVR

## Dependencies

### DIFF
--- a/Assets/GothicVR/Scripts/Creator/VobCreator.cs
+++ b/Assets/GothicVR/Scripts/Creator/VobCreator.cs
@@ -394,8 +394,8 @@ namespace GVR.Creator
 
             if (FeatureFlags.I.vobItemsDynamicAttach)
             {
-            grabComp.useDynamicAttach = true;
-            grabComp.selectMode = InteractableSelectMode.Multiple;
+                grabComp.useDynamicAttach = true;
+                grabComp.selectMode = InteractableSelectMode.Multiple;
             }
 
             var eventComp = vobObj.GetComponent<ItemGrabInteractable>();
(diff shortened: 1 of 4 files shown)


Assets/GothicVR/Scripts/Debugging/FeatureFlags.cs
```diff
@@ -27,6 +27,8 @@ public enum SunMovementPerformance
         public bool CreateWaypoints;
         [Tooltip("Leave blank if you want to spawn normal.")]
         public string spawnAtSpecificFreePoint;
+        [Tooltip("True will render all pickables with dynamic attach points")]
+        public bool dynamicGrab;
```
JaXt0r, line 31: Let's call it something like vobItemsDynamicAttach as it is more clear what it is for.

Assets/GothicVR/Scripts/Creator/VobCreator.cs
```diff
@@ -391,6 +391,12 @@ private static GameObject CreateItem(PxVobItemData vob)
             // It will set some default values for collider and grabbing now.
             // Adding it now is easier than putting it on a prefab and updating it at runtime (as grabbing didn't work this way out-of-the-box).
             var grabComp = vobObj.AddComponent<XRGrabInteractable>();
+
+            if(FeatureFlags.I.dynamicGrab){
```
JaXt0r, line 395: Browser coding.😉
Please reformat code. We use the C# style of brackets on next line.

JaXt0r (changes requested):
Some smaller changes. Please also add your setting with "true" within this class:
https://github.com/GothicVRProject/GothicVR/blob/main/Assets/GothicVR/Editor/Scripts/Tools/FeatureFlagTool.cs

It handles FeatureFlags during release Pipeline run. And I think your implementation is worth to be active for next release.

JaXt0r (approved):
Works pretty well. And yes, it's better for immersion to have flexible grabs in VR.

PR #266, opened 2023-11-14
Use given samplerate and channel number for Sounds

JucanAndreiDaniel:
Depends on https://github.com/GothicKit/phoenix-csharp-interface/pull/41

# To test
1. Walk around the eastern forest (near the sect camp) and see if bird chirping is still weird

# Testing
* [x] Tested with PCVR


JucanAndreiDaniel:
> Way better than before! Nice one.
> 
> I found these exceptions during gameplay, but I think it's unrelated to your change? image

yes, unrelated

PR #268, opened 2023-11-16
Routine waypoint based pathfinding POC

BoroBongo:
This change implements a global access to the FindFastestPath method in the DijkstraPathFinder static class. This allows any part of the code* to find the fastest path between two waypoints using Dijkstra's algorithm.

🚨 *  **Earliest it can be called is right after the WaynetCreator runs and Waypoint GameObject have been created**  🚨

# To test
1. While debugging with breakpoints - recreate steps in this video:
   [Watch the video](https://youtu.be/ywAUUVUPz_w)

# Checklist
1. Make Bloodwyn calculate one of his routines and start patroling ( Hopefully a big difference in immersion )

## Scene

# Testing
* [ ] Tested with PCVR

## Dependencies


Assets/GothicVR/Scripts/Creator/WaynetCreator.cs
```diff
@@ -11,10 +11,13 @@ public static class WaynetCreator
         public static void Create(GameObject root, WorldData world)
         {
             var waynetObj = new GameObject(string.Format("Waynet"));
-            waynetObj.transform.parent = root.transform;
 
+            DijkstraPathFinder pathFinder = waynetObj.AddComponent<DijkstraPathFinder>();
+            waynetObj.transform.parent = root.transform;
 
             CreateWaypoints(waynetObj, world);
+            pathFinder.SetDijkstraWaypointsOriginal(world); // Has to be here to get the waypoints position
```
BoroBongo, line 19: This can be moved to somewhere else BUT it can't be before the above CreateWaypoints method runs. I think this is the place for this so that any class at any time can use pathfinding. Happy to be proven wrong though!

Assets/GothicVR/Scripts/World/DijkstraPathFinder.cs
```diff
@@ -0,0 +1,308 @@
...
+            if (start == null)
+            {
+                start = DijkstraWaypoints[this.start] ?? throw new ArgumentException("The specified start waypoint is not in the queue.");
+            }
+            if (end == null)
+            {
+                end = DijkstraWaypoints[this.end] ?? throw new ArgumentException("The specified end waypoint is not in the queue.");
+            }
+
+            var startDijkstraWaypoint = DijkstraWaypoints[start];
```
Frozenenenen, line 160: Error here. You try to get the waypoint object with the key, which would have to be a string for this struct, but is actually a waypointObject.

Assets/GothicVR/Scripts/Npc/RoutePlanner.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace GVR
+{
+    public class RoutePlanner : MonoBehaviour
+    {
+        public List<PxWayPointData> CreatePath(string start, string end)
+        {
+            return null;
```
JaXt0r, line 12: Why this method, if it won't return anything?

If you wanted to start the route planner here, I suggest you use 
Assets/GothicVR/Scripts/Manager/WayNetHelper.cs
as it is already intended to be used for WayNet related topics at runtime.

Assets/GothicVR/Scripts/World/DijkstraWaypoint.cs
```diff
@@ -0,0 +1,35 @@
...
+using UnityEngine;
+using UnityEngine.InputSystem.LowLevel;
+
+namespace GVR
+{
+    public class DijkstraWaypoint : IComparable<DijkstraWaypoint>
+    {
+        public string _name = "";                                //as index to find other data like position, underwater and probably isFree
+
+        public double _coveredDistance = 0;
```
JaXt0r, line 13: Please follow coding style of C#.

If you use Rider, it will tell you, that underscore is for private (static?) only.
https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions

Assets/GothicVR/Scripts/World/DijkstraWaypoint.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.InputSystem.LowLevel;
+
+namespace GVR
+{
+    public class DijkstraWaypoint : IComparable<DijkstraWaypoint>
```
JaXt0r, line 9: Please add a few comments. Like
* Why is coveredDistance 0, why summedDistance 999
* You could also think of making this a Struct instead of class, as it contains only fields/properties.

Assets/GothicVR/Scripts/World/DijkstraWayPointCreator.cs
JaXt0r, line 1: Please move this function into
Assets/GothicVR/Scripts/Creator/WaynetCreator.cs

Whis is intended for Waynet related topics.

Assets/GothicVR/Scripts/World/DijkstraWayPointCreator.cs
```diff
@@ -0,0 +1,54 @@
...
+namespace GVR
+{
+    public class DijkstraWayPointCreator : MonoBehaviour
+    {
+        /// <summary>
+        /// A dictionary containing all the DijkstraWaypoints in the scene, with their names as keys.
+        /// </summary>
+        public static Dictionary<string, DijkstraWaypoint> DijkstraWaypoints;
+
+        public static Dictionary<string, DijkstraWaypoint> Create(WorldData world)
```
JaXt0r, line 17: Please add some comments what you're doing here and why.

Assets/GothicVR/Scripts/World/DijkstraWayPointCreator.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            Dictionary<string, DijkstraWaypoint> DijkstraWaypoints = new();
+            var wayEdges = world.waypointEdges;
+            var wayPoints = world.waypoints;
+    
+            foreach (var edge in wayEdges)
```
JaXt0r, line 23: Cool cool. If you want to flex, you could also refactor with LINQ. But just a bonus idea. ;-)

Assets/GothicVR/Scripts/World/DijkstraPathFinder.cs
```diff
@@ -0,0 +1,328 @@
...
+using System.Collections;
+using GVR.Vob.WayNet;
+using System.Linq;
+using System;
+using System.IO;
+
+namespace GVR
+{
+
+    public class DijkstraPathFinder : MonoBehaviour
```
JaXt0r, line 16: I would put this to Assets/GothicVR/Scripts/Manager/WaynetHelper.cs tbh.

I also didn't check for the code itself. I believe in you. :-D

It would be awesome, if we could create this path inside the Lab. This way we would have a proper integration test.
I can explain Lab to you these days.

Assets/GothicVR/Scripts/Creator/WorldCreator.cs
```diff
@@ -42,6 +42,7 @@ public static async Task CreateAsync(string worldName)
             await WorldMeshCreator.CreateAsync(world, teleportGo, ConstantsManager.MeshPerFrame);
             await VobCreator.CreateAsync(teleportGo, nonTeleportGo, world, ConstantsManager.VObPerFrame);
             WaynetCreator.Create(worldGo, world);
+            /*DijkstraWayPointCreator.Create(world);      //Is there a better place for this call? this might still work  */
```
JaXt0r, line 45: To be removed?

JaXt0r (changes requested):
A few code related hints. (I didn't tell you before, so now is the time ;-) )
I would also highly recommend to create a proper integration test. Let's leverage Lab for it.

Assets/GothicVR/Scripts/World/DijkstraWaypoint.cs
BoroBongo: changes made

Assets/GothicVR/Scripts/World/DijkstraWayPointCreator.cs
BoroBongo: adressed

Assets/GothicVR/Scripts/World/DijkstraWaypoint.cs
BoroBongo: added comments
when I change it to struct this is what I get
image

Assets/GothicVR/Scripts/World/DijkstraWayPointCreator.cs
BoroBongo: flexing with LINQ 💪

Assets/GothicVR/Scripts/World/DijkstraWayPointCreator.cs
BoroBongo: added comments

Assets/GothicVR/Scripts/World/DijkstraWayPointCreator.cs
BoroBongo: moved the method to waynetcreator

Assets/GothicVR/Scripts/Npc/RoutePlanner.cs
BoroBongo: @Frozenenenen what was your original idea for this?

Assets/GothicVR/Scripts/Creator/WaynetCreator.cs
```diff
@@ -11,13 +13,44 @@ public static class WaynetCreator
         public static void Create(GameObject root, WorldData world)
         {
             var waynetObj = new GameObject(string.Format("Waynet"));
-            waynetObj.transform.parent = root.transform;
 
+            DijkstraPathFinder pathFinder = waynetObj.AddComponent<DijkstraPathFinder>();
+            waynetObj.transform.parent = root.transform;
```
JaXt0r, line 18: I created a convenience function for this:
```suggestion
            waynetObj.SetParent(root);
```

Assets/GothicVR/Scripts/World/DijkstraPathFinder.cs
JaXt0r: In the WaynetHelper.cs class, we also work with static class. This is due to the fact, that we only use MonoBehaviour (GO-Components), if needed.
Based on our knowledge: The more we put into a scene, the more merge conflicts we create. So the helper/creator, ... are mostly static classes if no Unity magic like Update() etc. is needed.

Assets/GothicVR/Scripts/World/DijkstraPathFinder.cs
```diff
@@ -0,0 +1,439 @@
...
+                unvisited.Enqueue(waypointx, waypointx.SummedDistance);
+                // Add the waypoint to the previousNodes dictionary and set its previous node to null
+                previousNodes[waypointx.Name] = null;
+            }
+
+            while (unvisited.Count > 0)
+            {
+                var currentWaypoint = unvisited.Dequeue();
+                Debug.Log(currentWaypoint.Name + " " + currentWaypoint.SummedDistance);
+                LightUpWaypoint(currentWaypoint.Name, Color.yellow);
```
JaXt0r, line 197: I'd recommend you put these debug elements:
* remove them or
* hide them behind a feature flag

Your path finding algorithm is performance critical. Every last CPU cycle we can save, we should do so. ;-)

Assets/GothicVR/Scripts/World/DijkstraPathFinder.cs
```diff
@@ -0,0 +1,439 @@
...
+                SiftUp(index);
+            }
+            // If the new priority is greater than the old priority, sift down
+            else if (priority > oldPriority)
+            {
+                SiftDown(index);
+            }
+        }
+
+        //  "Sift" in the context of a heap data structure refers to the process of adjusting
```
JaXt0r, line 316: Thank you for a lot of comments. They'll help in the future.

Assets/GothicVR/Scripts/World/DijkstraWaypoint.cs
```diff
@@ -0,0 +1,36 @@
...
+        public Vector3 Position = new Vector3();  // Stores the position of the node
+
+        public string FatherWP = "";  // Stores the previous node in the shortest path
+
+        public DijkstraWaypoint(string name)
+        {
+            Name = name;
+        }
+
+        public int CompareTo(DijkstraWaypoint other)
```
JaXt0r, line 31: Why do you inherit IComparable<> if you don't want it to be compared? Feel free to remove it.

JaXt0r (changes requested):

Assets/GothicVR/Scripts/World/DijkstraPathFinder.cs
JaXt0r: Updated.

Assets/GothicVR/Scripts/World/DijkstraPathFinder.cs
JaXt0r: Updated.

Assets/GothicVR/Scripts/World/DijkstraWaypoint.cs
JaXt0r: Updated.

Assets/GothicVR/Scripts/Npc/RoutePlanner.cs
JaXt0r: Updated.

JaXt0r:
I integrated the path finder into existing WaynetCreator.cs / WaynetHelper.cs.

Still need to test the Dijkstra*Debug.cs method and see if the calculations are still there.

JaXt0r (approved):
I tested it, reworked your code into existing classes, and added an additional DebugComponent and put it below FeatureFlags/Debug

@Frozenenenen / @BoroBongo - You have the honor to hit the merge button. ⏯️ 🙇

JaXt0r:
I already tested it and use it in another branch. I merge it now.

PR #269, opened 2023-11-17
Also don't set the water as occluder

ThomasMountainborn:
Little thing I noticed that was missing, the water was also still set to be a static occluder, which it shouldn't be of course. I don't think there was an issue with it, but at least now we're sure there won't be.


Assets/GothicVR/Scenes/Worlds/world.zen/OcclusionCullingData.asset
JaXt0r, line 1: Do we have water in other worlds as well? If any, please check and update the file as well.

JaXt0r (changes requested):
one request.

JaXt0r:
At one point or another, we will update the OC data. No need to wait for it. Merging now.

JaXt0r:
Oh no merge conflict. I'll solve it soon.

JaXt0r (approved):

JaXt0r:
Lol. GitHub couldn't do auto-merge, but SourceTree did. Merged...

PR #272, opened 2023-11-27
Feature/270 root motion + Routines for WPs + grounded walk

JaXt0r:
# To test
1. Set FeatureFlag.spawnNPcs = 233 (Bloodwyn)
2. Set FeatureFlag.time = 8:55
3. Check if he walks to a FreePoint, waiting for a few seconds, then walking to a path to the other side of the camp

Boundary:
* Walk is still a little bit wonky, but Bloodwyn shouldn't fall through the ground.
* A lot of changes are due to rework of FeatureFlags

# Checklist

## Scene

# Testing
* [x] Tested with PCVR

## Dependencies


JaXt0r:
In review for 2 weeks+. I already forked from it and doing the migration to ZenKit. I will therefore merge it, as it is already used.

I found one bug: Bloodwyn is falling through the ground when walking through OC. But it's igkorable for now.

PR #287, opened 2024-01-10
Bugfix/Font-wrap-wrong

JucanAndreiDaniel:
Fix glyph metrics to have the glyphs with the right orientation and fix all labels from main menu

# To test
1. Open main menu and check if text looks the same 
2. Create a label with a lot of text so that it wraps and check if text orientation is fine

### DIFF
--- a/Assets/GothicVR/Resources/Prefabs/UI Element Prefabs/MainMenu.prefab
+++ b/Assets/GothicVR/Resources/Prefabs/UI Element Prefabs/MainMenu.prefab
@@ -100,13 +100,13 @@ RectTransform:
   m_PrefabInstance: {fileID: 0}
   m_PrefabAsset: {fileID: 0}
   m_GameObject: {fileID: 28688052990566104}
-  m_LocalRotation: {x: 1, y: 0, z: 0, w: 0}
+  m_LocalRotation: {x: 0, y: 0, z: 0, w: 1}
   m_LocalPosition: {x: 0, y: 0, z: -0}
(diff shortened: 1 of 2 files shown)


JaXt0r:
I tested your feature. Is this the intended outcome? If you can share a before - after image, then I can test better.
image

Assets/GothicVR/Scripts/UI/FontManager.cs
```diff
@@ -49,13 +49,14 @@ public TMP_SpriteAsset LoadFont(string fontName)
...
-                        horizontalBearingY = h,
+                        height = -h,
+                        horizontalBearingY = 0,
                         horizontalBearingX = 0,
                         horizontalAdvance = w
                     },
                     index = (uint)i,
-                    sprite = newSprite
+                    sprite = newSprite,
+                    scale = -1
```
JaXt0r, line 59: Why a scale of -1? Sounds wrong tbh.

Assets/GothicVR/Resources/Prefabs/UI Element Prefabs/MainMenu.prefab
JaXt0r, line 1: Hm. Seems you flip about every text element. Why?

JaXt0r (changes requested):

JucanAndreiDaniel:
> I tested your feature. Is this the intended outcome? If you can share a before - after image, then I can test better. image

This is intended


before
image

after

image


these are taken from the discord where i posted what was the issue

JaXt0r (approved):

PR #291, opened 2024-01-13
Feature/barrier

JucanAndreiDaniel:
# To test
1. Make sure that Barrier Flag is set
2. Start World and check if barrier exists
3. Compare our barrier with original game if it's the same waviness

# Testing
* [ ] Tested with PCVR


Assets/GothicVR/Scripts/Manager/BarrierManager.cs
```diff
@@ -0,0 +1,138 @@
...
+
+
+        public void CreateBarrier()
+        {
+            var barrierMesh = AssetCache.TryGetMsh("MAGICFRONTIER_OUT.MSH");
+            barrier = MeshObjectCreator.CreateBarrier("Barrier", barrierMesh, Vector3.zero, Quaternion.identity)
+                .GetAllDirectChildren()[0];
+        }
+
+        public void FixedUpdate()
```
JaXt0r, line 45: Please add a comment why we need it at FixedUpdate. As you subscribe to gameTimeChange listener, you could also leverage this one to start/stop timed.

Assets/GothicVR/Shaders/Barrier.shader
JaXt0r, line 1: Your first shader. Congratulations. 😃

Assets/GothicVR/Scenes/Bootstrap.unity
```diff
@@ -1614,6 +1616,7 @@ GameObject:
   - component: {fileID: 1896514223}
   - component: {fileID: 1896514226}
   - component: {fileID: 1896514228}
```
JaXt0r, line 1618: Please change _SkyManager.cs_ to _static_. You can then subscribe to GeneralSceneLoaded to load the actual sky. (Where you can also check for world name to load only for world.zen).

Background: We want to have as few SingletonBehaviours in our Bootstrap.unity scene as possible. As it is complex to read and adds potential merge conflicts and renaming issues in the future.

Assets/GothicVR/Editor/Scripts/Tools/FeatureFlagTool.cs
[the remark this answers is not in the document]
JucanAndreiDaniel: Since we directly modify the material when it is loaded as a skybox, unity doesn't return it to it's default values after play mode is deactivated

Assets/GothicVR/Scenes/Bootstrap.unity
JucanAndreiDaniel: I would rather not have SkyManager as a static class as it's not used as a helper class to provide only functions to be used, but rather having the sky state available for interpolation

Assets/GothicVR/Scripts/Manager/BarrierManager.cs
JucanAndreiDaniel: I wanted a consistent way to check if barrier needs to be faded in/out, and for now the fade in/out state is dependent on how fast we call the function, using fixed update we get a consistent fade in/out

JaXt0r:
Hi @JucanAndreiDaniel Are you still working on this one? Or shall I review again?

JucanAndreiDaniel:
> Hi @JucanAndreiDaniel Are you still working on this one? Or shall I review again?

Done with feature dev. need to sort out/refactor meshcreator, as i'm not happy with it's current state

JaXt0r (approved):
I aligned MeshCreator with other *MeshCreator.cs files. Tested and barrier works. We're good to go.

PR #293, opened 2024-01-20
Bugfix/fix adpcm decoding

JucanAndreiDaniel:
# To test
1. if bloodwyn speaks in lab
2. if text has the right rotation


JaXt0r (approved):
Thank you.

PR #297, opened 2024-01-26
Basic map feature

Romeeey:
# To test
1.) Press primary right
2.) Map should appear with animation
3.) Press it again. Map should dissapear
4.) Behaviour should be similar to menu

# Checklist
(Please remove entries which aren't needed for your PR)

## Scene

# Testing
* [ ] Tested with PCVR

## Dependencies
* [ ] If ZenKit.dll is changed, zenkitcapi.dll and libzenkitcapi.so are also changed


JaXt0r (approved):
Roll animation works fine.

PR #299, opened 2024-01-27
Bugfix/lab issues

JaXt0r:
# To test
1. Open Lab and walk to Bloodwyn
2. If Dialog triggers, then we're fine

Before the fix:
* Lab crashed after 2nd frame.


JucanAndreiDaniel (approved):
No crashes, LGTM

PR #303, opened 2024-01-30
Bugfix/fix teleport ray collider

JucanAndreiDaniel:
# To test
1. When spawning go a bit behind and see if teleport ray interacts with zonemusic

# Testing

* [ ] Tested with PCVR


JaXt0r (changes requested):
I updated the ignore layer. It should now work with:
* FP
* WP
* ZS

Please test it on your device if you can teleport onto all of these.

JaXt0r:
Suggestion from @ThomasMountainborn - let's move these elements from collider to trigger. Then we don't need to change the layer and physic based magic isn't calculated each frame any longer.

JucanAndreiDaniel:
> Suggestion from @ThomasMountainborn - let's move these elements from collider to trigger. Then we don't need to change the layer and physic based magic isn't calculated each frame any longer.

They are already triggers, it's just that the teleport raycast was set to collide with triggers

PR #302, opened 2024-01-29
Feature/more externals

JucanAndreiDaniel:
Adding more externals
For animation we need to properly test if it works

PR #304, opened 2024-01-31
improve(Climb): When on top of ladder, we now spawn towards ZS_POS1.

JaXt0r:
# To test
1. Load lab and climb on top of a ladder - You should be teleported where ZS_POS1 is.
2. Load world.zen and test the same. Every ladder should teleport you on top, when you reach the ZS_POS1 y-position.

Boundary:
* Teleporting via Raycaster doesn't work all the time. Tested and seems to be an issue from earlier.
* If we blink or teleport smoothly towards the position is a topic for later (e.g. when tester/gamer demands it).


JucanAndreiDaniel (approved):
Works just fine in lab, but in world only with [021fcd0](https://github.com/GothicVRProject/GothicVR/pull/302/commits/021fcd0eeb793066dde582feeb1c4172682122c9) applied

PR #306, opened 2024-02-07
Initial lighting

ThomasMountainborn:
Initial lighting. Supported now:
- Non-static light vobs
- Fire vobs are always on and not yet animated

# To test
1. Performance around the map

# Checklist
(Please remove entries which aren't needed for your PR)

# Testing
* [ ] Tested with PCVR


Assets/GothicVR/Scripts/Creator/WorldCreator.cs
```diff
@@ -25,7 +25,7 @@ public static class WorldCreator
         private static GameObject _worldGo;
         private static GameObject _teleportGo;
         private static GameObject _nonTeleportGo;
-        private static HashSet<IPolygon> _claimedPolygons;
+        private static readonly HashSet<IPolygon> ClaimedPolygons;
```
JucanAndreiDaniel, line 28: ClaimedPolygons is never initialized and so when trying to check if the hashset contains polygon, we get an error
Please Fix

JucanAndreiDaniel (changes requested):

Assets/GothicVR/Scripts/Creator/WorldCreator.cs
JaXt0r, line 28:
```suggestion
        private static readonly HashSet<IPolygon> ClaimedPolygons = new();
```

JaXt0r (approved):

JucanAndreiDaniel (approved):

JaXt0r (approved):
World switching works now.

PR #318, opened 2024-03-24
317 update pico openxr package

Tyroonis:
# Checklist
- Check apk on Pico

## Scene

# Testing
* [x] Tested with Pico

I have already tested it with my Pico.
It runs much more stable and nicely at 90FPS.

All buttons/functions work


JaXt0r (approved):
You're the pico man. If you tested it, it's fine. Nice we finally got rid of a preview version. 🤣

PR #320, opened 2024-03-24
bug fix - no physics on released items

piotrb92:
this is a fix to bug reported on discord here -> https://discord.com/channels/1000261958668406874/1221494186935586866

removed alreadyGrabbedOnce bool and checking if rigidbody isKinematic instead. only setting the physics on if they are off right now. assigning rigidbody on start instead of using getComponent on item release
moved VobMeshCullingManager.I.StopTrackVobPositionUpdates(gameObject); to the bottom of the release method sometimes there would be an error with vob mesh culling and that would stop physics from being re-enabled on item release

fixed checks in VobMeshCullingManager to make sure it never adds the same game object to the dictionary


Tyroonis:
works for me on PCVR and Pico

JaXt0r (approved):
Gameplay tested by @Tyroonis . Code looks fine. Go for it. :-)

PR #327, opened 2024-03-26
bugfix(RootCollisionHandler): Ignore colliders that are not freepoints

JucanAndreiDaniel:
Some waypoints begin with FP_ when they are waypoints on the waynet, so we just ignore them as they don't have vobSpotProperties

## To Test
1. change in Scripts/Manager/GvrSceneManager.cs line 71 from ```GameData.GothicVm.Call("STARTUP_SUB_OLDCAMP");``` to ```GameData.GothicVm.Call("STARTUP_WORLD");```
2. see if errors are thrown regarding rootcollisionhandler


JaXt0r (approved):
Thanks.

PR #328, opened 2024-03-26
227-SeatableObjects

piotrb92:
Added camera fade canvas with black image under player's camera in player prefab
 added CreateSeat method in VobCreator
created a folder under scripts for interactive objects 
added Seat script that handles sitting down and standing up 
added a folder under scripts->player for camera scripts 
added CameraFade script that handles camera fading 
all objects that contain "chair","bench" or "throne" are now seats that player can sit on by grabbing it, then stand up by grabbing it again.

# To test
1. sit on a bench
2. sit on a chair
3. sit on a throne
4. spam sitting and standing up as fast as possible


Assets/GothicVR/Scripts/Creator/VobCreator.cs
```diff
@@ -222,6 +223,16 @@ private static GameObject LoadVob(IVirtualObject vob, GameObject parent = null)
...
                 case VirtualObjectType.oCMobInter:
+                    {
+                        if(vob.Name.ContainsIgnoreCase("bench") ||
+                            vob.Name.ContainsIgnoreCase("chair") ||
+                            vob.Name.ContainsIgnoreCase("throne"))
+                        {
+                            go = CreateSeat(vob, parent);
+                            _cullingVobObjects.Add(go);
+                        }
+                        break;
```
JucanAndreiDaniel, line 234: We need to create a default mesh if it's not a seat

JucanAndreiDaniel (changes requested):

Assets/GothicVR/Scripts/Creator/VobCreator.cs
piotrb92: of course! thanks for checking :)

Assets/GothicVR/Scripts/Interactive/Seat.cs
```diff
@@ -0,0 +1,129 @@
...
+            {
+                float distance = Vector3.Distance(pos, snapPoints[i].position);
+                if (distance < minDistance)
+                {
+                    minDistance = distance;
+                    nearestSnapPoint = snapPoints[i];
+                }
+            }
+            return nearestSnapPoint;
+        }
```
JaXt0r, line 127: You could even use LINQ:
```suggestion
        private Transform GetNearestSnapPoint(Vector3 pos)
        {
            // Find the closest Snap Point for player to sit.
            return snapPoints
                .OrderBy(point => Vector3.Distance(pos, point.position))
                .First();
        }
```
Beautiful isn't it. :-D

Assets/GothicVR/Scripts/Player/Camera/CameraFade.cs
```diff
@@ -0,0 +1,35 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace GVR
```
JaXt0r, line 6:
```suggestion
namespace GVR.Player.Camera
```

Assets/GothicVR/Scripts/Player/Camera/CameraFade.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    public class CameraFade : MonoBehaviour
```
JaXt0r, line 8: Let's rework this a little, please.
As this seems to be a Singleton, we have two ways of handling this:
1. It's a static class, if we have no Unity calls at all (e.g. no Start or wired properties at Editor time)
2. Use a SingletonBehaviour<*> --> As you wire an image at editor time, I think we will use this.

SingletonBehaviour:
Look at e.g. FeatureFlags. They're also a SingletonBehavriour. In the end it's just a convenient wrapper to ensure:
1. We have an instance created only once in our scene
2. We have a fast lookup: e.g. FeatureFlags.I.*

Assets/GothicVR/Scripts/Player/Camera/CameraFade.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        public Image cameraFadeImage;
+
+        private void Start()
+        {
+            Fade(0.1f, 0);
```
JaXt0r, line 14: Magic number.
Can you make this 0.1f a const variable on top of the class, please. Easier to update if we need it in the future.

Assets/GothicVR/Scripts/Player/Camera/CameraFade.cs
```diff
@@ -0,0 +1,35 @@
...
+            StartCoroutine(FadeCamera(duration, targetAlpha));
+        }
+
+        private IEnumerator FadeCamera(float duration, float targetAlpha)
+        {
+            float currentTime = 0;
+
+            while (currentTime < duration)
+            {
+                currentTime += Time.fixedDeltaTime;
```
JaXt0r, line 28: I think you created an issue here.
Coroutines will be called every frame. Then you should use Time.deltaTime to set value every frame.

But! If you want to use Fixed (physics based?) updates, then you can optimize like this (I read on the forum):
```c#
currentTime += Time.fixedDeltaTime;
... // do your magic
yield return WaitForFixedUpdate(); // Now coroutine gets called again when we really want it. Ensures we skip unneccesary frames.
```

JaXt0r (changes requested):
That's a fine feature you built there. You optimized calls and cached important data already.
I added a few change requests and explained more in depth why. I hope it helps you understand a few decisions better.

Didn't test your feature so far. Will do either today or after your changes.

JaXt0r:
I can't sneak into the barrel. 😭 This is a major missing point. 😄 Gamers will test this first!
image

Assets/GothicVR/Scripts/Player/Camera/CameraFade.cs
piotrb92: pretty sure it was supposed to be deltaTime, dont know how fixed got in there xD

Assets/GothicVR/Scripts/Player/Camera/CameraFade.cs
piotrb92: used SingletonBehaviour<*>

Assets/GothicVR/Scripts/Interactive/Seat.cs
piotrb92: fair shout!

piotrb92:
all requested changes implemented

Assets/GothicVR/Scripts/Interactive/Seat.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        private void Start()
+        {
+            gameObject.layer = LayerMask.NameToLayer("Interactive"); //set layer to interactive so we can interact using XR Ray interactor
+            //get snap points
+            GetSnapPoints();
+            //get interactable
+            interactable = GetComponent<XRGrabInteractable>();
+            //add ToggleSitting listener to SelectEntered
+            interactable.selectEntered.AddListener(ToggleSitting);
```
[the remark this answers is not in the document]
piotrb92: it's on start, so only when the object is created, not every time it gets enabled.

piotrb92 (commented):
all requested changes implemented

JaXt0r (changes requested):
Thanks for the updates. A few remarks:
* From my point of view, the positions are too high. Look a the barrel, it seems I'm flying above it. Benches are similar. (I was in a seated position and the normal height of floor was ok.)
* I can't lie in beds. There's no marker to hop onto.
* (There was one time where rotation didn't activate again after unselecting a seat. But I couldn't reproduce (just for the record))

Additional ideas (possibly for later):
* Currently we select the items by selecting the ZS_* collider. In original G1, you select the whole object. The ZS_* is just the animation starting point. (Feel free to put this in a separate ticket if you don't want to do it now)
* Hoping of a seating animation is looking down and selecting the bench again. It's a little unnatural behaviour. I think a button is better suited for this. (But I think you can leave this for now until we come up with proper button mappings)

piotrb92:
@JaXt0r if this is a photo of you being in the barrel, then you need to re-center your headset. This is not how it's supposed to look.
I think beds should be a separate feature, because you can sleep in a bed, so they are different than seats.
I don't really understand what you mean by the last 2 points, so we might have to talk about it

PR #329, opened 2024-03-27
Add initial implementation of zCVobAnimate

JucanAndreiDaniel:
Add initial implementation of zCVobAnimate


Assets/GothicVR/Scripts/Caches/AssetCache.cs
```diff
@@ -196,7 +196,15 @@ public static IMorphMesh TryGetMmb(string key)
             if (MmbCache.TryGetValue(preparedKey, out var data))
                 return data;
 
-            var newData = new MorphMesh(GameData.Vfs, $"{preparedKey}.mmb").Cache();
+            IMorphMesh newData = null;
+            try
```
JaXt0r, line 200: Oh. Don't tell me there are object again, where the mmb file doesn't exist? :-)

JaXt0r (changes requested):
Unfortunately I couldn't find an object which has this animation.

I also don't like we introduce certain MorphMesh parameters for PrepareMeshFilter(), yes, some of them are legacy of mine, but we need to be clear here.

I will check it during Easter and try to optimize.

Assets/GothicVR/Scripts/Caches/AssetCache.cs
JucanAndreiDaniel: just in case 😉

JaXt0r (commented):
Hey hey. I changed your wobble grass branch a little.
* The complains about AbstractMeshBuilder were more on my side. I improved the parameters a little.
* You were kind of lazy, my friend. :-D HeadMorph contains VobAnimate logic... Tststs. :rofl:  - I created a subclass for both.

Please check briefly and once ok for you, you can merge.

JaXt0r (approved):

PR #331, opened 2024-03-29
Bugfix/loading errors

JaXt0r:
# To test
1. Chest in Diego's house is rotated right.
2. There are only 4 errors in log about "unknown Gothic version for world. Assuming G1" (Before that: ~140)
3. No log error about Culling


Assets/GothicVR/Scripts/Creator/VobCreator.cs
```diff
@@ -272,9 +278,12 @@ private static GameObject CreateFire(Fire vob, GameObject parent = null)
             if (vob.VobTree == "")
                 return go;
 
-            var vobTree = new ZenKit.World(GameData.Vfs, vob.VobTree);
+            if (!vobTreeCache.TryGetValue(vob.VobTree.ToLower(), out IWorld vobTree))
+            {
+                vobTree = new ZenKit.World(GameData.Vfs, vob.VobTree).Cache();
```
JucanAndreiDaniel, line 283:
```suggestion
                vobTree = new ZenKit.World(GameData.Vfs, vob.VobTree);
```

JucanAndreiDaniel (changes requested):

JaXt0r:
I'll also add, that we leverage latest ZenKit version which also solves the last error message.

Assets/GothicVR/Scripts/Creator/VobCreator.cs
JaXt0r: We can keep this cache. At least the .zen itself isn't re-parsed every time.

JaXt0r:
@JucanAndreiDaniel please test again. Unfortunately I couldn't see your fire again. But the errors are gone.
Please also check if the caching still makes sense. I think so, but 🤷

JucanAndreiDaniel (approved):
Make sure to restart unity as it might not let you see all the lights, i see no issue on my side

PR #333, opened 2024-04-01
Feature/npc idle dialogs

JaXt0r:
Merge blocked by: https://github.com/GothicVRProject/GothicVR/pull/310

# To test
1. Spawn Scar (101) and Arto (102)
2. Go into Gomez throne room
3. Listen to them talking

Bonus:
Please record a small audio for Romeyyy. He'll love it. ;-)

Boundary - can be ignored for now ;-) :
* NPCs don't fully look towards each other
* They sometimes slightly rotated away from other NPC before next talk
* If you spawn on a Freepoint where e.g. Arto spawns, it tries to kind of initialize a hero dialog, but fails. Therefore Arto is stuck in place. It's a topic for another PR, but for now: Spawn at a FP/WP where no NPC is right now, please.


Assets/GothicVR/Scripts/Npc/Actions/AnimationActions/UseItemToState.cs
```diff
@@ -1,55 +1,88 @@
...
 
 namespace GVR.Npc.Actions.AnimationActions
 {
     public class UseItemToState : AbstractAnimationAction
     {
         
-        private const string animationStartScheme = "T_{0}_STAND_2_S{1}" ;
-        private const string animationEndScheme = "T_{0}_S{1}_2_STAND" ;
-        
+        private const string animationScheme = "T_{0}_{1}_2_{2}";
```
Romeeey, line 12: What is that scheme? is that hardcoded or does it come from the original?

Assets/GothicVR/Scripts/Npc/Actions/AnimationActions/UseItemToState.cs
```diff
@@ -1,55 +1,88 @@
...
+
+
         public UseItemToState(AnimationAction action, GameObject npcGo) : base(action, npcGo)
         { }
 
         public override void Start()
         {
-            // Nothing in hand && new item shall be put into hand
-            if (Props.itemAnimationState < 0 && Action.Int1 >= 0)
+            PhysicsHelper.DisablePhysicsForNpc(Props);
```
Romeeey, line 24: Why is physics deactivated here? is that a hack / workaround or is there a proper reason for it?

Assets/GothicVR/Scripts/Npc/Actions/AnimationActions/UseItemToState.cs
```diff
@@ -1,55 +1,88 @@
...
+                newItemAnimationState = ++Props.itemAnimationState;
             }
-            // Something in hand && item shall be removed
-            else if (Props.itemAnimationState >= 0 && Action.Int1 < 0)
+            else
             {
-                EndItemStateAnimation(Props.itemAnimationState);
                 Props.hasItemEquipped = false;
+                Props.currentItem = -1;
+                // e.g. Babe brush doesn't call it automatically. We therefore need to force remove the brush item from hand.
```
Romeeey, line 42: That's pretty specific. How did you come up with that finding? Is that a bug / problem in the original game? Is it a proper solution?

Romeeey (approved):
Function test was successful. I submitted some questions, which would be interesting to be answered but overall the PR is good to go from my side.

JucanAndreiDaniel (changes requested):
Commit [e57a63d](https://github.com/GothicVRProject/GothicVR/pull/333/commits/e57a63d3380a3d73b91c56493797f2a256ee7c19) makes unity crash as there is a stack overflow in DaedalusVM probably because of missing hero instance being properly instanciated

JaXt0r:
> Commit [e57a63d](https://github.com/GothicVRProject/GothicVR/pull/333/commits/e57a63d3380a3d73b91c56493797f2a256ee7c19) makes unity crash as there is a stack overflow in DaedalusVM probably because of missing hero instance being properly instanciated

Oh my. What did you do? Any specific FeatureFlags you used etc.? I will try to reproduce to understand...

Assets/GothicVR/Scripts/Npc/Actions/AnimationActions/UseItemToState.cs
JaXt0r: It's from Gothic. I rent it from OG and also double checked with Gothic animation files. The names are visible inside loaded .anim assets from ZK.

JaXt0r: Documentation is something for the future. ;-)

Assets/GothicVR/Scripts/Npc/Actions/AnimationActions/UseItemToState.cs
JaXt0r: Good point. I added a comment:
```c#
            // We need to disable physics as e.g. Gomez won't sit on his throne otherwise. (Animations are aligned with the objects they use already).
            // A collider would break it.
```

JaXt0r:
I fixed the hero lookup. Please check again.

JucanAndreiDaniel (approved):
go for it :D

PR #334, opened 2024-04-04
Update Zenkit to latest version

JucanAndreiDaniel:
Fixes color issues for r5b6g5 textures --> look at the sky. Is it ok again?


Romeeey (approved):
Functionality check is ok.

PR #347, opened 2024-04-23
Physics disable while interactions + UseMob fix for Gomez and Cooks + Walk fixes

JaXt0r:
# To test
1. Boot game with the following NPCs:
2. Buddler (527) - The one on the water. He should wash himself and shouldn't stutter (aka physics are disabled during washing)
3. Buddler (502) - He should use cauldron again to cook.
4. Gomez (100) - He will now directly sit on the throne without walking away first.
5. Guard (214) - He should guard now. (Was walking into the wall before.


BoroBongo (approved):
Nice work! LGTM :D

PR #349, opened 2024-04-24
fix(GothicSky.mat): Changes during Editor.OnValidate() removed as the…

JaXt0r:
…y will be set at game startup without saving to .mat file directly.

Two major changes:
1. I removed the OnValidate() function which calls ```SetShaderProperties()``` as this alters values inside GothicSky.mat before the game starts but should only reflect runtime changes. The ```SetShaderProperties()``` is also called during game startup which is:
```WorldCreator.CreateAsync() -> InitSky() -> Interpolate() -> InterpolateSky() -> SetShaderProperties()``` after world is loaded and before loading screne is gone.
2. GothicSky was set directly on world.unity scene. Therefore we couldn't overwrite it. Now setting the value at runtime when world is fully loaded dynamically as a clone.

Tests showed the sky is fine at startup.

### DIFF
--- a/Assets/GothicVR/Scenes/Worlds/world.zen.unity
+++ b/Assets/GothicVR/Scenes/Worlds/world.zen.unity
@@ -26,7 +26,7 @@ RenderSettings:
   m_AmbientIntensity: 1
   m_AmbientMode: 3
   m_SubtractiveShadowColor: {r: 0.42, g: 0.478, b: 0.627, a: 1}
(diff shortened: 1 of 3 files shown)


ThomasMountainborn (approved):
Downside of removing OnValidate is that changing light colors in the inspector only takes place when the in-game clock moves forward, i.e. not when paused and not when loading the world in the editor. But that's not necessarily a big problem. We want the light color values to be driven by the sky manager anyway. So it's fine by me.

PR #351, opened 2024-04-28
Bugfix/lab errors

JaXt0r:
# To test
1. Start Lab scene
2. Walk into Bloodwyn
3. You shouldn't see any error on the log and his dialog should proceed until the dialog popup opens
(Before that the dialog stopped after hero said something)


BoroBongo (approved):
Very nice! 😎

PR #356, opened 2024-05-06
NPC breathing and blinking

JaXt0r:
# To test
1. Load Lab scene
2. Activate GVRSingletons --> FeatureFlags --> enableNpcEyeBlinking
3. Spawn an NPC and play his animation
4. After the animation you should see NPC blinking (too long, but it works already) and breathing

Boundary:
* The feature for breathing won't really work in main game as NPCs are always doing something.
* Blinking is behind a feature flag so that we can pause implementation and rework in the future.

image


Redreaper666 (approved):
EyeBlink and Breathing works as intended and can be merged.

PR #363, opened 2024-05-23
Update unity to 2022.3.30f

JucanAndreiDaniel:
Update the project to use unity 2022.3.30f.
Fixes debug mode on quest2/3 so we can finally profile on mobile.

# Testing
* [x] Tested with PCVR


JucanAndreiDaniel:
> Is it latest LTE?

latest LTS released 2 days ago. seems that they are releasing a new version every week, but it fixes the only issue i had with the current version we use: debug and profiling on mobile

JaXt0r (approved):

